PR #2, opened 2025-12-17
Play with `Extractor` / `Segment` architecture

lauraporta:
As mentioned in #1, me and @alessandrofelder have been playing around with the spikeinterface `Extractor` / `Segment` architecture and built a module for suite2p registration, inheriting from `BasePreprocessor` and `BasePreprocessorSegment`. 
@alejoe91, would be happy to know what do you think!


alejoe91:
Hi guys! Great start :)

Motion correction is a tricky one though, let me try to explain why. 
Motion correction/registration includes 2 steps:
- motion estimation
- video interpolation

In your current implementation, suite2p both steps are in the segment `get_series` function. The problem with this is that if a user calls the function with different frame slices (e.g., `get_series(100, 200)`/`get_series(1000, 2000)`/`get_series(0, 10000)`) they will get different results. 
I think that a better implementation (that we also have in spikeinterface) is the following:
- have a `compute_motion()` function (see the same function in spikeinterface [here](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/motion.py#L282)): this function traverses the entire video and computes rigid/non-rigid motion signals for the entire duration. This step is therefore **not lazy**
- have an `InterpolateImaging` preprocessor, that takes in the motion signals and applies them on the fly in the `get_series` function (this is the spikeinterface [class](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/sortingcomponents/motion/motion_interpolation.py#L276)). 

This makes sure that whatever the slicing, users will always get the correct registration throughout the full duration.

Does it make sense?

alessandrofelder:
I am happy with your suggested implementation (IIUC it's often a slow step anyway, so we'd probably want to precompute and write to file anyway).

>user calls the function with different frame slices (e.g., get_series(100, 200)/get_series(1000, 2000)/get_series(0, 10000)) they will get different results.

Out of curiousity and to improve my understanding - I am a bit of novice still when it comes to multiphoton imaging and `suite2p` - why would this get different results (assuming we use the same registration target, e.g. an average of the first 150 frames of the video, not the requested series)?

alejoe91:
RIght, I see the logic now. Yes if you use the first 150 as target you would probably get the same results! my bad :)

NOt sure if we want to enforce this, let's discuss later!

lauraporta:
Closing this PR as registration features are now developed in #27

PR #11, opened 2025-12-23
feat: add registration based on Alessio's feedback

arielleleon:
@alejoe91 @lauraporta 

- Added `compute_motion_suite2p` to calculate displacements
- Displacements applied when `RegisterSuite2PImaging` is instantiated


arielleleon:
I realize this isn't as pressing as our other items but wanted to push anyway to save the work

PR #13, opened 2026-01-19
Add core classes and extensive tests

### LINKED ISSUE #5: base classes with tests
This will be a cleaner version of parts of #2 

This issue will be addressed when
* we have `BaseImaging`, `BaseSegment`, `BaseROIs` classes
* they have minimal tests which will cover
  * serialisation/deserialisation
  * writing/reading to binary

From our developer call, we agree to use `pytest` and aim for good, but not perfect test coverage.

alejoe91:
closes #5

- Added base classes in core modules: `BaseImaging`, `BaseRois`, `BinaryImaging`, `NumpyImaging`, `NumpyRois`.
- Added support for multi-plane volumetric data
- Added extensive tests of core module

TODO:

- [x] Fix mypi annotations


src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,278 @@
...
+    sampling_frequency : float
+        The sampling frequency
+    image_shape : tuple(int, int)
+        Image height and width
+    dtype : str or dtype
+        The dtype of the binary file
+    num_planes : int, default: 1
+        Number of planes in the imaging data
+    plane_ids : list[int] | None, default: None
+        List of plane IDs. If None, defaults to [0, 1, ..., num
```
alessandrofelder:
```suggestion
        List of plane IDs. If None, defaults to [0, 1, ..., num_planes]
```
I think?

src/photon_mosaic/core/numpyimaging.py
```diff
@@ -0,0 +1,198 @@
...
+
+        num_segments = len(videos)
+        self._sampling_frequency = float(sampling_frequency)
+
+        # Check that all shapes and number of planes are consistent across segments
+        shapes = []
+        for video in videos:
+            if len(video.shape) not in [3, 4]:
+                raise ValueError(
+                    "'timeseries' must be a 3D or 4D numpy array (num_frames, height, width, [num_channels])"
```
alessandrofelder:
```suggestion
                    "'timeseries' must be a 3D or 4D numpy array (num_frames, height, width, [num_planes])"
```
I think?

src/photon_mosaic/core/numpyimaging.py
```diff
@@ -0,0 +1,198 @@
...
+
+import numpy as np
+
+from .baseimaging import BaseImaging, BaseImagingSegment
+from .baserois import BaseRois
+from .utils import ArrayType, FloatType
+
+
+class NumpyImaging(BaseImaging):
+    """An single-segment Imaging specified by timeseries .npy or numpy array"""
```
alessandrofelder: I may have missed it, but where is the functionality that allows reading from `.npy` files 🤔

src/photon_mosaic/core/generators.py
```diff
@@ -0,0 +1,92 @@
+"""Module to generate synthetic imaging and ROI objects for testing purposes."""
```
alessandrofelder: If these are purely for testing purposes, should they be moved [to a `conftest.py` file in the test folder and be made `pytest.fixtures`](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)?

src/photon_mosaic/core/tests/test_baserois.py
```diff
@@ -0,0 +1,162 @@
...
+from photon_mosaic.core.baserois import BaseRois
+from photon_mosaic.core.generators import generate_rois
+
+
+class MockImaging:
+    def __init__(self, sampling_frequency):
+        self.sampling_frequency = sampling_frequency
+
+
+def _make_rois_for_tests() -> tuple[BaseRois, dict]:
```
alessandrofelder: should this be a local `pytest.fixture`?

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,227 @@
...
+            The pixel coordinates for the specified ROIs.
+        """
+        if roi_ids is None:
+            roi_ids = self.roi_ids.tolist()
+
+        # Get pixel masks from representations
+        pixel_masks = []
+        image_masks = self.get_roi_image_masks(roi_ids)
+        for img_mask in image_masks:
+            # 2D case
```
alessandrofelder: Should we make an issue for the 3D case so we don't forget?

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,278 @@
...
+            pass
+
+
+# For backward compatibility (old good time)
+read_binary = BinaryImaging
+
+
+class BinaryFolderImaging(BinaryImaging):
+    """
+    BinaryFolderImaging is an internal format used in spikeinterface.
```
alessandrofelder:
```suggestion
    BinaryFolderImaging is an internal format used in photon-mosaic.
```
Presumably?

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,227 @@
...
+        Returns
+        -------
+        SelectRois
+            A new BaseRois object containing only the selected ROIs.
+        """
+        return SelectRois(self, roi_ids)
+
+    def register_imaging(self, imaging: BaseImaging):
+        """
+        Register an imaging to the sorting. If the sorting and imaging both contain
```
alessandrofelder:
```suggestion
        Register an imaging to the ROIs. If the ROIs and imaging both contain
```
I think?

alessandrofelder (commented):
Thanks a lot for this @alejoe91 - this was easy to review thanks to the clean code and our previous experiments and discussions.

I have some nitpicky comments (some may be linked to me misunderstanding the code) but overall I'd be happy to approve this once you've considered them.

There are a number of TODOs in the code - would it be better to move those out to issues so everything is in one place? (I can help with this boring work 😁 )

If we decide to go down the `dask` rather than the `Segment` design route, we can do a major refactoring (#14) so I am happy for that choice to not be a blocker for now.

src/photon_mosaic/core/generators.py
alejoe91: I think these can be used in examples and tutorials as well, so let's keep them here :)

src/photon_mosaic/core/baserois.py
alejoe91: good point, I can directly extend and test for the 3d case

src/photon_mosaic/core/numpyimaging.py
alejoe91: you're right, I modified it so it's only arrays

alejoe91:
```suggestion
    """An single-segment or multi-segment Imaging specified by a numpy array or list of arrays"""
```

alejoe91:
> Thanks a lot for this @alejoe91 - this was easy to review thanks to the clean code and our previous experiments and discussions.
> 
Thanks for reviewing! I addressed most comment :)

> I have some nitpicky comments (some may be linked to me misunderstanding the code) but overall I'd be happy to approve this once you've considered them.
> 
> There are a number of TODOs in the code - would it be better to move those out to issues so everything is in one place? (I can help with this boring work 😁 )
> 
It would be highly appreciated!

> If we decide to go down the `dask` rather than the `Segment` design route, we can do a major refactoring (#14) so I am happy for that choice to not be a blocker for now.

I vote to keep segments for 2 reasons:
1. without segments, we would not be able to inherit from the Base spikeinterface object, so we would need to re-implement a lot of functionalities
2. I think that in some cases the segment abstraction is useful, like in a session with several not continuous acquisitions.

We could still add a dask backend directly on the SpikeInterface side though if needed!

src/photon_mosaic/core/tests/test_baserois.py
alejoe91: done!

src/photon_mosaic/core/baserois.py
alejoe91: done!

alejoe91:
@alessandrofelder all comments should be addressed now

src/photon_mosaic/core/generators.py
alessandrofelder:
```suggestion
"""Module to generate synthetic imaging and ROI objects for testing and example purposes."""
```

alessandrofelder: Nitpicky, but I think helps readers understand why this is not a fixture?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+import numpy as np
+from spikeinterface.core.base import BaseExtractor
+from spikeinterface.core.chunkable import ChunkableMixin, ChunkableSegment
+from spikeinterface.core.chunkable_tools import get_chunks, write_binary
+from spikeinterface.core.core_tools import convert_bytes_to_str, convert_seconds_to_str
+
+from photon_mosaic.core.utils import DtypeType
+
+# TODO: frames instead of samples
+# TODO: epoch instead of segment (segmentation is another thing)
```
alessandrofelder:
```suggestion
```
Moved to https://github.com/photon-mosaic/photon-mosaic/issues/15

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+
+import numpy as np
+from spikeinterface.core.base import BaseExtractor
+from spikeinterface.core.chunkable import ChunkableMixin, ChunkableSegment
+from spikeinterface.core.chunkable_tools import get_chunks, write_binary
+from spikeinterface.core.core_tools import convert_bytes_to_str, convert_seconds_to_str
+
+from photon_mosaic.core.utils import DtypeType
+
+# TODO: frames instead of samples
```
alessandrofelder:
```suggestion
```
Moved out to https://github.com/photon-mosaic/photon-mosaic/issues/16

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+            binary_imaging.dump(folder / "binary.json", relative_to=folder)
+
+            cached = BinaryFolderImaging(folder_path=folder)
+
+        elif format == "memory":
+            raise NotImplementedError
+        elif format == "zarr":
+            raise NotImplementedError
+        elif format == "nwb":
+            # TODO implement a format based on zarr
```
alessandrofelder:
```suggestion
```
zarr #17 and NWB #18 issues created

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+import mmap
+import warnings
+from pathlib import Path
+
+import numpy as np
+
+from .baseimaging import BaseImaging, BaseImagingSegment
+
+
+# TODO: make it more flexible in terms of specifying axes
```
alessandrofelder: I am not entirely certain what you mean here - so maybe you could write this one up into an issue @alejoe91 ?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+            folder = kwargs["folder"]
+            file_paths = [folder / f"video_cached_seg{i}.raw" for i in range(self.get_num_segments())]
+            dtype = kwargs.get("dtype", None) or self.get_dtype()
+            t_starts = self._get_t_starts()
+
+            write_binary(self, file_paths=file_paths, dtype=dtype, verbose=verbose, **job_kwargs)
+
+            from .binaryimaging import BinaryFolderImaging, BinaryImaging
+
+            # This is created so it can be saved as json because the `BinaryFolderRecording` requires it loading
```
alessandrofelder: Should this be `BinaryFolderImaging`? (also in docstring of `is_binary_compatible` and on line below)?

alessandrofelder (approved):
Thanks @alejoe91 

I've made a number of nitpicky comments for small improvements but I am approving this - which (in my world of conventions - we should document?) means I trust you to consider my comments (you can accept or reject them) and then merge without reviewers needing to have another look :) 

I will move the conversation about segments to #14

src/photon_mosaic/core/tests/test_baseimaging.py
```diff
@@ -0,0 +1,215 @@
+import numpy as np
```
lauraporta, line 1: Minor: 
shall we move the tests out of `src/` and keep them in `/tests/`?

src/photon_mosaic/core/utils.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
+
+import numpy as np
+from numpy.typing import ArrayLike, DTypeLike
+
+ArrayType = ArrayLike
```
lauraporta: Minor: I am not fully convinced of aliasing the `dtypes` as doesn't seem we gain that much in terms of character lengths and they will be unfamiliar to external developers.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+    @property
+    def plane_ids(self):
+        """Get the plane IDs associated with the imaging data.
+
+        Returns
+        -------
+        list
+            A list of plane IDs.
+        """
+        return self._main_ids
```
lauraporta, line 135: Minor: 
I don't see `self._main_ids` being set anywhere so I guess it is a property that will be taken care by classes that inherit it. I wonder if we can set it to `0` in the constructor just to avoid being thrown an error.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        int
+            The total number of frames.
+        """
+        return self.get_num_samples(segment_index=segment_index)
```
lauraporta, line 230: General on Segments:
My understanding here is one `Segment` maps to one movie, so its `len` = `time`, at least for the conceptualisation of this PR.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+        -------
+        int
+            The total number of samples (frames).
+        """
+        if segment_index is None:
+            if self.get_num_segments() == 1:
+                segment_index = 0
+            else:
+                raise ValueError("segment_index must be provided for multi-segment imaging data.")
+        return self.segments[segment_index].get_num_samples()
```
lauraporta, line 207: General on Segments:
Here I am guessing, the `List` of `Segments` is a list of single movies (corresponding to different trials if I merge the understanding before) coming from an unique tiff. I guess we are in a one plane world
I see some overlap with `get_num_frames`, as for now it enforces only one segment and returning an int.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+    """
+    Abstract class representing a multichannel timeseries, or block of raw ephys traces
+    """
+
+    def get_series(
+        self,
+        start_frame: int,
+        end_frame: int,
+        plane_indices: list[int] | None = None,
+    ) -> np.ndarray:  # pragma: no cover
```
lauraporta, line 429: Why are we excluding `get_series()` from coverage?

lauraporta:
Hi, my review is in progress :) should be done by today!

.pre-commit-config.yaml
```diff
@@ -25,6 +25,8 @@ repos:
     -   id: mypy
         additional_dependencies: [types-setuptools]
         args: [--ignore-missing-imports]
+        exclude: ^.*/tests/.*\.py$
+
```
lauraporta, line 29: Why do we exclude tests?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+    Each `BaseImaging` instance is associated to a single "channel".
+    The `_main_ids` attribute is used here for multi-plane imaging objects.
+    """
+
+    def __init__(self, sampling_frequency: float, shape: tuple | list | np.ndarray, plane_ids: list | None = None):
+        if plane_ids is None:
+            plane_ids = [0]  # dummy single plane
+        BaseExtractor.__init__(self, plane_ids)
+        self._sampling_frequency = float(sampling_frequency)
+        assert len(shape) == 2, "Shape must be a tuple/list/array of length 2 (height, width)"
```
lauraporta: Minor:
I remember this `assert` becoming false every time I was playing around with volumetric data for the other dask related PR. The shape was never explicitly declared, but it was inferred  by the size of the tiff... Indeed it was passed in this way: `shape=self.roi_extractor.get_sample_shape()` 🤔 which fails when calling multiple-palne data.

General on Segments:
Besides this, I guess this `assert` stays in place as it supposes a one plane world.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+        self._image_shape = np.array(shape)
+        self._average_image = None
+
+    def _repr_header(self, display_name=True):
+        """Generate text representation of the BaseImaging object."""
+        num_samples = [self.get_num_samples(segment_index=i) for i in range(self.get_num_segments())]
```
lauraporta, line 35: General on Segments:
Just call `self.get_num_samples()`if it could return a `List[int]` for multiple planes?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+        """Get the number of pixels in the image.
+
+        Returns
+        -------
+        int
+            Number of pixels in the image.
+        """
+        return np.prod(self.image_shape)
+
+    def get_num_planes(self) -> int:
```
lauraporta, line 262: Wait, don't we have a property for this?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+        """Get the number of planes in the imaging data.
+
+        Returns
+        -------
+        int
+            The number of planes.
+        """
+        return len(self.plane_ids)
+
+    def get_sampling_frequency(self):
```
lauraporta, line 159: Why do we need a get method? let's just have the property public and call it.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+        self,
+        num_chunks: int = 20,
+        chunk_duration: str = "1s",
+        chunk_size: int | None = None,
+        recompute: bool = False,
+    ) -> np.ndarray:
+        if self._average_image is not None and not recompute:
+            return self._average_image
+        else:
+            data = get_chunks(
```
lauraporta, line 339: General on SpikeInterface:
Why is chunking not mediated through the Segment class here? When it's the native Segment chunking operation coming? is it calling a `get_series()` under the hood?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+
+        Parameters
+        ----------
+        imaging_segment : BaseImagingSegment
+            The imaging segment to add.
+        """
+        self.segments.append(imaging_segment)
+        imaging_segment.set_parent_extractor(self)
+
+    def is_binary_compatible(self) -> bool:
```
lauraporta, line 349: What does binary compatible mean?

src/photon_mosaic/core/baseimaging.py
lauraporta: I agree ☝️

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+                num_planes=self.get_num_planes(),
+                dtype=dtype,
+                t_starts=t_starts,
+                file_offset=0,
+            )
+            binary_imaging.dump(folder / "binary.json", relative_to=folder)
+
+            cached = BinaryFolderImaging(folder_path=folder)
+
+        elif format == "memory":
```
lauraporta, line 400: What is format "memory"?

src/photon_mosaic/core/baseimaging.py
```diff
@@ -0,0 +1,434 @@
...
+            # TODO implement a format based on zarr
+            raise NotImplementedError
+
+        else:
+            raise ValueError(f"format {format} not supported")
+
+        for segment_index in range(self.get_num_segments()):
+            if self.has_time_vector(segment_index):
+                # the use of get_times is preferred since timestamps are converted to array
+                time_vector = self.get_times(segment_index=segment_index)
```
lauraporta, line 413: General on SpikeInterface:
What is a time vector?

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,258 @@
+import numpy as np
+from numpy.typing import ArrayLike
+from spikeinterface.core.base import BaseExtractor
+
+from .baseimaging import BaseImaging
+
+
+class BaseRois(BaseExtractor):
```
lauraporta, line 8: Not many strong opinions about `BaseRois` as I'm very far from wrapping my head around it yet! 🤣

I wander were we will be having also an analogous `BaseNeuropil`, or if neuropil data will be joint to this class. But that's a disccussion in a few months time 😊

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,258 @@
...
+        """Get the pixel coordinates for a specific ROI.
+
+        Parameters
+        ----------
+        roi_ids : list[int | str] | None
+            The IDs of the ROIs.
+
+        Returns
+        -------
+        np.ndarray
```
lauraporta, line 166: If useful, it can be a float representing the probability of belonging to a ROI, at least based on my understanding of suite2p

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,258 @@
...
+            The IDs of the ROIs to select.
+
+        Returns
+        -------
+        SelectRois
+            A new BaseRois object containing only the selected ROIs.
+        """
+        return SelectRois(self, roi_ids)
+
+    def register_imaging(self, imaging: BaseImaging):
```
lauraporta, line 204: Is this just a reference to the pre-registered movie from which the ROIs are from?

src/photon_mosaic/core/baserois.py
```diff
@@ -0,0 +1,258 @@
...
+        assert np.isclose(
+            self.sampling_frequency, imaging.sampling_frequency, atol=0.1
+        ), "The imaging has a different sampling frequency than the ROIs!"
+        self._imaging = imaging
+
+
+class SelectRois(BaseRois):
+    """Class to select a subset of ROIs from an existing BaseRois object."""
+
+    def __init__(self, rois: BaseRois, roi_ids: ArrayLike):
```
lauraporta, line 227: I guess we can also add the probability of being a cell coming from the classifier

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+            # one segment
+            file_path_list = [Path(file_paths)]
+
+        if t_starts is not None:
+            assert len(t_starts) == len(file_path_list), "t_starts must be a list of the same size as file_paths"
+            t_starts = [float(t_start) for t_start in t_starts]
+
+        dtype = np.dtype(dtype)
+
+        for i, file_path in enumerate(file_path_list):
```
lauraporta, line 66: General on Segments:
OK so from here I can see clearly that the idea is that multiple segments are different trials in possibly one experimental session

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+        d = dict(
+            file_paths=self._kwargs["file_paths"],
+            dtype=np.dtype(self._kwargs["dtype"]),
+            image_shape=self._kwargs["image_shape"],
+            num_planes=self._kwargs["num_planes"],
+            file_offset=self._kwargs["file_offset"],
+        )
+        return d
+
+    def __del__(self):  # pragma: no cover
```
lauraporta, line 102: Wonder if there are alternative ways to delete the object? I am not a garbage collector expert, my understanding is that `__del__` might never by called if some refs remain 🤔

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+
+        # Adjust the length so it includes the extra data from rounding down
+        # the memmap offset to a multiple of ALLOCATIONGRANULARITY
+        length += start_offset
+
+        # Create the mmap object
+        memmap_obj = mmap.mmap(
+            self.file.fileno(),
+            length=length,
+            access=mmap.ACCESS_READ,
```
lauraporta, line 160: I guess this is what grants laziness, am I correct?

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+        series = np.ndarray(
+            shape=shape,
+            dtype=self.dtype,
+            buffer=memmap_obj,
+            offset=start_offset,
+        )
+
+        # Slice planes if needed
+        if plane_indices is not None and self.num_planes > 1:
+            series = series[:, :, :, plane_indices]
```
lauraporta: General on Segments:
What are we doing of this series with multiple planes?

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+        except Exception as e:
+            warnings.warn(f"Error closing file handle in BinaryImagingSegment: {e}")
+            pass
+
+
+# For backward compatibility (old good time)
+read_binary = BinaryImaging
+
+
+class BinaryFolderImaging(BinaryImaging):
```
lauraporta, line 204: I guess this is in relation to imaging object containing segments from multiple trials? so basically the imaging object already maps the folder in my understanding... in such a case why having an extra class with just metadata?

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+            raise ValueError("This folder is not a binary photon-mosaic folder")
+
+        assert d["relative_paths"]
+
+        d = make_paths_absolute(d, folder_path)
+
+        BinaryImaging.__init__(self, **d["kwargs"])
+
+        folder_metadata = folder_path
+        self.load_metadata_from_folder(folder_metadata)
```
lauraporta, line 239: General on SpikeInterface:
Is it expecting a certain metadata format? what is it? Guess this requires a further conversation on metadata, including tiff metadata.
Made an issue: #19

src/photon_mosaic/core/numpyimaging.py
```diff
@@ -0,0 +1,219 @@
+"""Imaging and Segmentation Extractors for .npy files.
```
lauraporta: How realistic do we want our synthetic data to be? I understand that a random numpy object it's ok to test basic functionalities in unit tests. I guess we will have different testing data complexity. I have made an issue #20

src/photon_mosaic/core/numpyimaging.py
```diff
@@ -0,0 +1,219 @@
...
+
+        num_segments = len(videos)
+        self._sampling_frequency = float(sampling_frequency)
+
+        # Check that all shapes and number of planes are consistent across segments
+        shapes = []
+        for video in videos:
+            if len(video.shape) not in [3, 4]:
+                raise ValueError(
+                    "'timeseries' must be a 3D or 4D numpy array (num_frames, height, width, [num_planes])"
```
lauraporta, line 59: General on Segments:
I can see here it is including the possibility of having segments with multiple planes

src/photon_mosaic/core/numpyimaging.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Returns
+        -------
+        series: np.ndarray
+            The raw series for the specified frame range.
+        """
+        start = start_frame if start_frame is not None else 0
+        end = end_frame if end_frame is not None else self._video.shape[0]
+        if self._video.ndim == 4 and plane_indices is not None:
+            return self._video[start:end, :, :, plane_indices]
```
lauraporta: I guess no chunking here, no laziness

src/photon_mosaic/core/tests/test_baseimaging.py
```diff
@@ -0,0 +1,215 @@
...
+    assert avg3.shape == (8, 6)
+    assert avg3 is not avg1
+
+
+# -------------------------
+# Multi-plane specific tests
+# -------------------------
+
+
+def test_multiplane_basic_shape_and_plane_ids():
```
lauraporta, line 167: I expected this to fail, but it does not as it does not expose the retreival of the shape itself from a slice of a tiff file (see one of the first comments on the assert).

lauraporta (approved):
Hi @alejoe91! Very cool to see the project coming along piece by piece! I've run the tests locally and they pass and I can see the CI are correctly picking them up.

I’ve left a number of comments, some of which are mostly a memo or a discussion point. Happy to know your thoughts on them. When a topic comes up repeatedly, comments are grouped under a short title.

I have marked some of them as *minor*, majority about class design. I can see that for certain data there is both a property and a get method and that confused me. I wonder if that is because of inheritance from SpikeInterface clsses.

Another portion of the comments concerns the concept of `Segment` itself, which we’ll revisit and discuss further beyond this PR. At the moment there seems to be a partial mix of code intended for multi-plane support, but not fully implemented yet. That’s probably fine for now, as these concepts will need another pass anyway. 
Edit: I can see also that @alessandrofelder has created an ~issue~ PR to discuss about sessions, so we can also reply to my comments about it in that conversation. 

I have marked a subset of other comments as General on SpikeInterface.

Thanks for pushing this forward! 🙌🏻

src/photon_mosaic/core/baseimaging.py
alejoe91: You would expect multi-plane objects to be a stack and have the same number of samples (basically a 4d array), no?

alessandrofelder:
Scope of this PR, as agreed in dev meeting
* Allow Imaging objects' `main_ids` to handle 3D/4D smoothly (by adding extra dimension to 3D data and avoid many re-occuring if/else clauses)
* Finalise current comments
* and https://github.com/photon-mosaic/photon-mosaic/pull/21 (could also be separate if you want @alejoe91)

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+        return d
+
+    def __del__(self):  # pragma: no cover
+        """
+        Ensures that all segment resources are properly cleaned up when this recording extractor is deleted.
+        Closes any open file handles in the recording segments.
+        """
+        # Close all recording segments
+        if hasattr(self, "_recording_segments"):
```
alejoe91: this should be `segments`

src/photon_mosaic/core/baseimaging.py
alejoe91: fixed

src/photon_mosaic/core/baseimaging.py
alejoe91: yes, get some frames is the same function but if we prefer using `frames` instead of samples

src/photon_mosaic/core/numpyimaging.py
alejoe91: The `get_series` will return the data in numpy, so when you call get_series it's when you break the laziness

src/photon_mosaic/core/baseimaging.py
alejoe91: yes these are set in the `BaseExtractor`

src/photon_mosaic/core/baseimaging.py
alejoe91: because this base one is never going to be called and is not implemented. Needs to be implemented in subclasses

.pre-commit-config.yaml
alejoe91: because it's honestly a pain to check all typing in tests...I undrestand enforcing it in the main src code, but it becomes a massive overload to also do it in tests, since we might want to be more flexible there

src/photon_mosaic/core/baseimaging.py
alejoe91: Some spikeinterface functions still rely on the method...I think it doesn;t hurt to have both!

src/photon_mosaic/core/baseimaging.py
alejoe91: this is sampling chunks across all segments! The definition of segments is that they are observing the same neural tissue, so it makes sense to get random chunks across all (internally it is of course mediated by segments)

src/photon_mosaic/core/baseimaging.py
alejoe91: It the imaging object is a binary file/representation.

src/photon_mosaic/core/baseimaging.py
alejoe91: Format `memory` means loading the entire object in memory, but this can be done in parallel.

Helpful for small-ish recordings and large RAM machines

src/photon_mosaic/core/baseimaging.py
alejoe91: time vector is the timestamps / vector of times in seconds associated to the video

src/photon_mosaic/core/baserois.py
alejoe91: this can be done as a property. I don't think it should be part of the base class, but we can discuss later

src/photon_mosaic/core/binaryimaging.py
alejoe91: we have this in spikeinterface and it helps clean up some warings ;)

src/photon_mosaic/core/binaryimaging.py
alejoe91: No. At this point we are in the `get_series` function, so we just want to read a piece of the video and return it in memory. We just do it with a memmap and the correct offset ;)

src/photon_mosaic/core/binaryimaging.py
alejoe91: Here with `metadata` we load additional properties that have been set to the Imaging object. Properties are by ID, so in this case it would be a propery of planes. It could be for example the z coordinate of the stack.

src/photon_mosaic/core/binaryimaging.py
alejoe91: I'll just remove it. The TODO meaning was to being able to specify how the axis (time, H, W, planes) are organized within the file. Since this is mainly internal, I think we can dropit for now. If we encounter the need, we can add this option later.

src/photon_mosaic/core/utils.py
alejoe91: I agree. Removed

alejoe91:
> Scope of this PR, as agreed in dev meeting
> 
> * Allow Imaging objects' `main_ids` to handle 3D/4D smoothly (by adding extra dimension to 3D data and avoid many re-occuring if/else clauses)

Done, also for `BaseRois`

> * Finalise current comments

Done!

> * and [Rename Segment to Epoch #21](https://github.com/photon-mosaic/photon-mosaic/issues/21) (could also be separate if you want @alejoe91)

I'll do it in a follow up!


@lauraporta @alessandrofelder @arielleleon I'll merge so we can build on top of this

src/photon_mosaic/core/numpyimaging.py
alejoe91: replied on the issue

src/photon_mosaic/core/baseimaging.py
lauraporta: Yes this can be an interesting addition as it frequent to contrast "time bases" vs "frame based" analysis of the movies in my experience

src/photon_mosaic/core/baseimaging.py
lauraporta: I see makes sense!

src/photon_mosaic/core/baseimaging.py
lauraporta: Yes makes total sense!

src/photon_mosaic/core/baseimaging.py
lauraporta: OK interesting. I am trying to understand the use case. We are using the `_save` function here, so it means that we are processing the whole dataset, and instead of saving it to disk we just keep it entirely in memory as it is small enough... basically bringing the whole array to a given state

src/photon_mosaic/core/baseimaging.py
lauraporta: Cool! very similar to an object I've also seen in `pynapple` project!

src/photon_mosaic/core/baserois.py
lauraporta: OK, I've not a formed opinion yet on the matter!

PR #14, opened 2026-01-20
Replace segments with dask arrays

lauraporta:
Here an experiment on how we could substitute the Segment class with `dask` arrays inside Imaging objects.


alessandrofelder:
Hey @alejoe91 and @arielleleon 

Laura and I have been doing some more thinking around the software design of photon mosaic, and asked for feedback from our wider RSE team.

One thing that came up is that I couldn’t fully justify the Segment part of the design during the discussion.

-My understanding is that
- Imaging objects delegate lazy slicing and IO of parts of the to their segment(s)
- The segment functionality is mostly private: other code should interact with the Imaging objects

But maybe I am missing something?

Arguments against it were
- Isn’t this what `dask` was made for?
- Are the Segment and the Imaging objects too tightly coupled? Why does the `BaseROIExtractorImaging` need to know about the `roiextractor` if the reading is done on the segment side?

Just as an experiment, Laura and I experimented a bit yesterday around replacing `BaseROIExtractorImagingSegment` with just a `dask` array.

Naively, this seems maybe a bit more flexible in how we might want to (re)chunk and might save us some lines to maintain (We might also be able to remove the dependency on `ChunkableMixin`). We could chunk a dask array that is a member of the Imaging object along time (and possibly z) - this would enable lazy access to the array data without having to implement it. We might also benefit from dask’s functionality to parallelise across chunks.

This idea would preserve the design of the Imaging object: the get_series() method would be still in use but the implementation will only be inside the Imaging object. There would still be a stack of calls to parent get_series() and the actual operations on the data will happen through the .compute() method. It can be used after: get_series().compute(), or in certain cases (such as for registration) .compute() has to be executed also internally inside get_series() (e.g. to compute the average image template to register to).

Sorry for being slow - it feels like it’s quite an important decision so we’d like everyone to be as happy as possible with the key data structures before building lots of things around it :)  Very interested in what you think about this?

(People were also unhappy with “Segment” as nomenclature - I think we should urgently change to something else to avoid confusing image analysis people)

alessandrofelder:
Quote by @alejoe91  from #13
>I vote to keep segments for 2 reasons:
> * without segments, we would not be able to inherit from the Base spikeinterface object, so we would need to re-implement a lot of functionalities
> * I think that in some cases the segment abstraction is useful, like in a session with several not continuous acquisitions.
> We could still add a dask backend directly on the SpikeInterface side though if needed!

I have some questions here
* in the argument about inheritance: are you saying we can't inherit from `BaseSegment` or `BaseImaging`?
* in the argument about non-continuous acquisitions: are you implying that this is the main purpose of segments you have in mind? Or even the only purpose?

The below is possibly rooted in my ignorance!

I am wondering whether a dask array underlying the imaging object (instead of segments) could just always have a planes and a trial axis, and often these would have size 1. AFAICT this has the advantages of
* less mental gymnastics
* simpler code: fewer if/else statements checking for 2D/3D case and for 1/multiple segments, fewer loops (just always iterate over axis of interest, which sometimes has size 1)
* still being able to benefit from the `BaseImaging` class's provenance chain and (de-)serialisation functionality (which I like)

I think the extra axes should come with neglegible performance loss.

I would welcome more in-depth explanation of the advantages of the segment concept, which could sway me towards them again - sorry for being slow :snail: !

alejoe91:
> Quote by @alejoe91 from #13
> 
> > I vote to keep segments for 2 reasons:
> > 
> > * without segments, we would not be able to inherit from the Base spikeinterface object, so we would need to re-implement a lot of functionalities
> > * I think that in some cases the segment abstraction is useful, like in a session with several not continuous acquisitions.
> >   We could still add a dask backend directly on the SpikeInterface side though if needed!
> 
> I have some questions here
> 
> * in the argument about inheritance: are you saying we can't inherit from `BaseSegment` or `BaseImaging`?
> * in the argument about non-continuous acquisitions: are you implying that this is the main purpose of segments you have in mind? Or even the only purpose?
> 

Yes that would be the main use case. But we could choose to get rid of the segments and only accept mono-segment objects.

> The below is possibly rooted in my ignorance!
> 
> I am wondering whether a dask array underlying the imaging object (instead of segments) could just always have a planes and a trial axis, and often these would have size 1. AFAICT this has the advantages of
> 

I don't think this would work, since trials / segments could have different number of samples.

> * less mental gymnastics
> * simpler code: fewer if/else statements checking for 2D/3D case and for 1/multiple segments, fewer loops (just always iterate over axis of interest, which sometimes has size 1)
> * still being able to benefit from the `BaseImaging` class's provenance chain and (de-)serialisation functionality (which I like)
> 
> I think the extra axes should come with neglegible performance loss.
> 
> I would welcome more in-depth explanation of the advantages of the segment concept, which could sway me towards them again - sorry for being slow 🐌 !

We are also trying to wrap our minds around the idea! Let's discuss in depth on Monday! Another possibility could be having the `BaseSegment` (`BaseEpoch`) be a dask array, which is very easy (I have a draft :))

alessandrofelder:
> Let's discuss in depth on Monday! Another possibility could be having the BaseSegment (BaseEpoch) be a dask array, which is very easy (I have a draft :))

Yea! :raised_hands: :tada:

lauraporta:
Hi, I have been reading the thread and my take is to agree first on the data we are supporting, by making explicit all the dimensions we will be dealing with. Then work around what can be the best data structure to contain it, in a way that maximises efficiency and usability. Either Segments or dask or a mix of the two, as long as we have looked through the challenges from a 360 point of view. 

I'd hopefully have some schematics for our next chat!

lauraporta:
Closing this PR as discussions have moved away from an architecture without Segments (now renamed as Epochs).

PR #24, opened 2026-02-05
Standardize imaging arrays to always use 4D shape (T, H, W, P)

lauraporta:
Me and @alessandrofelder worked on the following.

### Changes
- **Breaking change**: All imaging data now consistently uses 4D arrays with shape `(num_frames, height, width, num_planes)`
- Removed `plane_ids` parameter. Plane IDs auto-generated as `range(num_planes)` to maintain consistency with `BaseExtractor` constructor
- `image_shape` now includes plane dimension: `(height, width, num_planes)` instead of `(height, width)`

### Open question: Terminology from neuroscience/microscopy perspective
**Is an "image" 2D or 3D?** From a microscopy background, should we consider:
- Current approach: `image_shape` includes planes (treating image as volumetric)
- Alternative: Separate `image_shape` (H, W) from `volume_shape` (H, W, P)?
- Planes get special treatment with dedicated property and attribute... should width and height have as well?

### Limitations
- **ROI classes unchanged**: Deferring 2D vs 3D segmentation decisions 🙈
- **Binary imaging slicing**: Currently slices by time dimension with memmap. Slicing by planes would be more efficient but requires more complex offset calculations... possible future work with Dask! 🤞🏻

### DIFF
--- a/src/photon_mosaic/core/baseimaging.py
+++ b/src/photon_mosaic/core/baseimaging.py
@@ -23,13 +23,15 @@ class BaseImaging(BaseExtractor, ChunkableMixin):
     The `_main_ids` attribute is used here for multi-plane imaging objects.
     """
 
-    def __init__(self, sampling_frequency: float, shape: tuple | list | np.ndarray, plane_ids: list | None = None):
-        if plane_ids is None:
-            plane_ids = [0]  # dummy single plane
-        BaseExtractor.__init__(self, plane_ids)
+    def __init__(self, sampling_frequency: float, shape: tuple | list | np.ndarray):
+        # Should we allow users to provide 2D shape (H, W) for single plane imaging?
+        if len(shape) == 2:
+            shape = (shape[0], shape[1], 1)
+        assert len(shape) == 3, "Shape must be a tuple/list/array of length 3 (height, width, planes)"
+
+        BaseExtractor.__init__(self, range(0, shape[2]))
         self._sampling_frequency = float(sampling_frequency)
-        assert len(shape) == 2, "Shape must be a tuple/list/array of length 2 (height, width)"
-        self._image_shape = np.array(shape)
+        self._image_shape = np.array(shape) # Image is intended as a volume (H, W, planes)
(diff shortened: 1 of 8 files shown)


src/photon_mosaic/core/baseimaging.py
```diff
@@ -23,13 +23,15 @@ class BaseImaging(BaseExtractor, ChunkableMixin):
...
     """
 
-    def __init__(self, sampling_frequency: float, shape: tuple | list | np.ndarray, plane_ids: list | None = None):
-        if plane_ids is None:
-            plane_ids = [0]  # dummy single plane
-        BaseExtractor.__init__(self, plane_ids)
+    def __init__(self, sampling_frequency: float, shape: tuple | list | np.ndarray):
+        # Should we allow users to provide 2D shape (H, W) for single plane imaging?
+        if len(shape) == 2:
+            shape = (shape[0], shape[1], 1)
```
alejoe91, line 29: Yes I think that's fair. Also usere will never instantiate this base object, so it would be mainly on child classes to set this correctly

src/photon_mosaic/core/baseimaging.py
```diff
@@ -23,13 +23,15 @@ class BaseImaging(BaseExtractor, ChunkableMixin):
...
+        assert len(shape) == 3, "Shape must be a tuple/list/array of length 3 (height, width, planes)"
+
+        BaseExtractor.__init__(self, range(0, shape[2]))
         self._sampling_frequency = float(sampling_frequency)
-        assert len(shape) == 2, "Shape must be a tuple/list/array of length 2 (height, width)"
-        self._image_shape = np.array(shape)
+        self._image_shape = np.array(shape) # Image is intended as a volume (H, W, planes)
```
alejoe91, line 34:
```suggestion
        self._shape = np.array(shape)
```
Let's just go with `_shape`

alejoe91 (approved):
Thanks @lauraporta 

I'll merge on core and make some small changes there!

PR #25, opened 2026-02-09
`segments` -> `epochs` (where possible) and `samples` -> `frames`

### LINKED ISSUE #15: Rename `segment` to `epoch` or `trial`
This issue will be addressed when
* we've agreed on what segments represent in `photon-mosaic`
* we've agreed on a nomenclature that disambiguates this from the word "segment"=="create a segmentation".
* we've refactored our docstrings and function names accordingly

alejoe91:
Some `segment_index` need to stay for SpikeInterface cmpatibility, but high level functions can use epochs and frames (preferred to segment/sample)

Fixes #15


alessandrofelder (approved):
Looks good to me!

PR #27, opened 2026-02-11
Registration feedback

arielleleon:
- Integrated the teams changes from the dev branch getting suite2p to work
- exposed suite2p and caiman settings to the user
- example suite2p registration notebook added to examples


src/photon_mosaic/core/utils.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
```
alejoe91: We got rif of utils in core and using direct typing

src/photon_mosaic/extractors/__init__.py
```diff
@@ -0,0 +1,38 @@
+# Import base classes
+from .roiextractors import (
```
alejoe91: @arielleleon can you push the extractors module in a separate PR?

src/photon_mosaic/preprocessing/basepreprocessor.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
```
alejoe91, line 1: I think we can only add the `basepreprocessor` class and remove the edge detector/lablace/median filters. Registration could be a submodule of preprocessing, since it is essentially a preprocessing step

src/photon_mosaic/segmentation/run_segmentation.py
```diff
@@ -0,0 +1,29 @@
+from photon_mosaic.core import BaseImaging, BaseRois
```
alejoe91: Let's add segmentation in a separate PR, since we probably want to target lower level functions

src/photon_mosaic/widgets/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .series import plot_imaging_series
```
alejoe91: Same for widgets :) let's keep it separate

examples/demo.ipynb
```diff
@@ -0,0 +1,341 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
```
alejoe91, line 4: Can you clear the notebooks? Otherwise they get quite heavy in git history

src/photon_mosaic/registration/baseregistration.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from photon_mosaic.core import BaseImaging, BaseImagingEpoch
+
+
+class BaseRegistration(BaseImaging):
+    def __init__(self, imaging, sampling_frequency=None, dtype=None):
```
alejoe91: Couldn't this be a `BasePreprocessor`? IMO registration could be a submodule of preprocessing

src/photon_mosaic/preprocessing/suite2p.py
```diff
@@ -0,0 +1,237 @@
...
+        Additional keyword arguments
+    """
+
+    def __init__(self, parent_imaging_epoch, motion, epoch_index, **kwargs):
+        BaseRegistrationEpoch.__init__(self, parent_imaging_epoch)
+        self.motion = motion
+        self.epoch_index = epoch_index
+        self.kwargs = kwargs
+
+    def get_series(self, start_frame, end_frame, plane_indices=None):
```
alejoe91, line 128: Why can't we use the `displacement_per_epochs` here? I think that if we want to use the suite2p function, we just need to pre-compute the `refAndMasks`, not the entire displacement. Then the `get_sereis` could compute and apply the displacement on the fly.

The alternative solution would be to pre-compute displacements per epochs, but then not rely on suite2p: in this case we would need to reimplement the interpolation (which could be re-used by Caimann as well)

alejoe91 (changes requested):
A few comments @arielleleon

src/photon_mosaic/extractors/__init__.py
arielleleon: Done

src/photon_mosaic/widgets/__init__.py
arielleleon: Done

src/photon_mosaic/segmentation/run_segmentation.py
arielleleon: Done

src/photon_mosaic/core/utils.py
arielleleon: Removed

src/photon_mosaic/preprocessing/basepreprocessor.py
arielleleon: I added it to a separate registration library so that it would be obvious for developers to know where to go. I will move it preprocessor and am happy to discuss at another time

src/photon_mosaic/preprocessing/suite2p.py
arielleleon: Take a look at the last commit and let me know what you think

arielleleon:
Closing - @seanmcculloch  is working on this

PR #30, opened 2026-02-12
28 add extractor to dev

### LINKED ISSUE #28: Add extractor to dev
Add the extractor library to the `dev` branch

arielleleon:
- Adds ROIExtractor to handle various input formats.


src/photon_mosaic/extractors/roiextractors.py
```diff
@@ -0,0 +1,210 @@
...
+        return self.roiextractor_extractor.get_num_samples()
+
+    def get_series(self, start_frame=None, end_frame=None, plane_indices=None):
+        series = self.roiextractor_extractor.get_series(start_frame, end_frame)
+        if plane_indices is not None and series.ndim > 3:
+            series = series[..., plane_indices]
+        return series
+
+
+def get_imaging_extractor(file_path: str | Path, imaging_name: str | None = None, **kwargs) -> BaseROIExtractorImaging:
```
alejoe91: Do we actuially need this function? Do we expect users to use it when all the `read_*` functions are available?

alejoe91 (changes requested):
This looks great @arielleleon 

My only question is whether we really need the `get_imaging_extractor`?

src/photon_mosaic/extractors/roiextractors.py
```diff
@@ -0,0 +1,210 @@
...
+    """Base class for ROI extractors that work with BaseImaging data."""
+
+    def __init__(self, roi_extractor_imaging: ImagingExtractor):
+        BaseImagingEpoch.__init__(self, sampling_frequency=roi_extractor_imaging.get_sampling_frequency())  # type: ignore[call-arg]
+        self.roiextractor_extractor = roi_extractor_imaging
+
+    def get_num_samples(self):
+        return self.roiextractor_extractor.get_num_samples()
+
+    def get_series(self, start_frame=None, end_frame=None, plane_indices=None):
```
lauraporta, line 43: As we've moved to 4D arrays, I think in here we've to add some logic to deal with shape == 3 (only one plane) and add an empty axis as it is not handled in the parent class.

src/photon_mosaic/extractors/tests/test_roiextractors.py
```diff
@@ -0,0 +1,192 @@
...
+    """Create a mock ImagingExtractor from roiextractors."""
+    mock = MagicMock()
+    mock.get_num_samples.return_value = num_samples
+    mock.get_sample_shape.return_value = shape
+    mock.get_sampling_frequency.return_value = sampling_frequency
+    mock.get_series.return_value = np.random.default_rng(0).random((num_samples, *shape))
+    return mock
+
+
+# --------------- BaseROIExtractorImagingEpoch ---------------
```
lauraporta, line 26: It would be very useful to include testing for all different roiextractor classes that we always can wrap it in a 4D array, shape is as expected and validate the input type as memmap or numpy array. This would clarify also the doubts of the diversity in return types of this package 😄

lauraporta (commented):
Hi @arielleleon, thanks for moving forward with this PR!
I left two comments generally on the need of interfacing with roiextractor different shapes and datatypes. I am happy to contribute to the branch as well with the changes required. 

I was thinking also, I didn't even begin of thinking properly how to deal with suite2p outputs. Shall we just not merge the ROIs related functionality for now and get there once we sort registration and segmentation? At that point we would be in a better position to test these classess functionality.

src/photon_mosaic/extractors/roiextractors.py
arielleleon: Good catch - should be fixed

src/photon_mosaic/extractors/tests/test_roiextractors.py
arielleleon: take a look now :)

src/photon_mosaic/extractors/roiextractors.py
arielleleon: Should be resolved

alejoe91 (approved):

Issue #51, opened 2026-03-19
Benchmarking reading and writing from arrays (`numpy memmap`; `zarr`; `dask`; `binary`)


lauraporta:
## Reporting progresses in benchmarking movie storage formats for multiphoton data

In the [`benchmark-dask`](https://github.com/photon-mosaic/photon-mosaic/tree/benchmark-dask) branch, we implemented a benchmark suite to compare storage/access backends for multiphoton imaging movies under realistic constraints:

- datasets may not fit in RAM (need out-of-core/lazy access),
- analysis often reads only subsets (small frame windows + selected planes).

### What we benchmarked
- `npy_full_load` (small-data baseline)
- `npy_memmap`
- `npy_dask` (dask over memmap)
- `binary_memmap`
- `zarr_dask` (native on-disk chunks)
- `zarr_dask_rechunked` (misaligned read chunks)
- `zarr_native` (direct zarr slicing, no dask)

### Benchmark grid
- planes on disk: 1 vs 50
- queried planes: 1 plane vs all planes
- start frame: aligned start vs mid-file seek
- frames per read: 10 vs 500
- dataset size: small (1000 frames) and large (4000 frames)

This gives us a reproducible way to see where each backend is most performant, and to make a data-driven choice for a default format and chunking strategy.

![Benchmark plots](https://raw.githubusercontent.com/photon-mosaic/photon-mosaic/refs/heads/benchmark-dask/benchmark_plots/bench_combined.png)

As you can see from the generated plots, `numpy memmap` is the winner in the majority of the cases. `Zarr + dask` wins when we want to access only one plane in a large volume. 

Next step is now merging the current changes in `dev`. In particular we are interested in content from #34 PR and benchmark the Spikeinterface-native functions to read and write zarr against the current implementations in the branch.

Issue #50, opened 2026-03-19
`ZarrImaging` class does not take `sampling_frequency` as constructor argument


alessandrofelder:
In #34, we introduce the `ZarrImaging` class.

It introduces a small breakage in the constructor API, because it doesn't take `sampling_frequency` as an argument.
It instead seems to [read the sampling frequency from zarr group metadata](https://github.com/SpikeInterface/spikeinterface/blob/f632aa7ccfb55309f45adde14a6091963f5860dc/src/spikeinterface/core/zarrextractors.py#L16).

We (@lauraporta and I) are not sure how to best handle this. Perhaps naively, we think we should be able to impose this on the zarr Group, and keep the Imaging classes' initialisation API consistent.

alejoe91:
Why is this a breaking change? The `sampling_frequency` is only exposed in the `BaseImaging`. Inherited classes are supposed to read/retrieve this from files (e.g., proprietary formats) and set it automatically. Or am I missing something?

alessandrofelder:
Ah - I was missing something then. I (and some of our draft tests) assumed that we could rely all `Imaging` objects following `BaseImaging.__init__`'s signature, but it makes sense to
* generally read it from standard file formats
* be able to set it via the constructor in tests or special cases like `NumpyImaging` where it cannot be gotten from file. 

I can make a PR to clarify this in the `BaseImaging` docstring - thanks @alejoe91

PR #52, opened 2026-03-19
(Do not merge) a place to discuss the benchmarks guiding our datastructure design.

alessandrofelder:
Following @lauraporta's work on benchmarking the IO of different array backends in situations we expect might happen in multi-photon data processing (e.g. reading a single plane from a volumetric imaging) (see #51), we have today incorporated a benchmark for the `ZarrImaging` class from latest `dev`.

I ran the benchmarks on by Ubuntu desktop, and the plot it added to this PR.
The `ZarrImaging` object from Laura's previous work was kept for comparison, and renamed to `ZarrImaging_Laura` for clarity.

Opening this PR to mainly discuss two things:
* why does our current ZarrImaging perform significantly worse than the other zarr-based approaches?
* (relatedly?) could this be because of how we hack together a single-epoch zarr Group in the `_get_or_create_zarr_group` helper function?

This could also be another kind of bug in the test code, for sure.

We also discovered https://github.com/photon-mosaic/photon-mosaic/issues/50 while working on this.


alejoe91:
From a very first look, it seems that it behaves worse in the benchmark with multiple planes, asking for a single plane. This is because by default chunks are only in time. We can easily change the chunking scheme to chunk on planes too!

alejoe91:
> From a very first look, it seems that it behaves worse in the benchmark with multiple planes, asking for a single plane. This is because by default chunks are only in time. We can easily change the chunking scheme to chunk on planes too!

Nevermind, I see that both use the same chunking...I'll keep exploring

alejoe91:
Performance of Zarr dev should be fixed here: #53

alessandrofelder:
Fixed in #53 :tada:

Issue #46, opened 2026-03-16
Utility to merge Imaging objects


alessandrofelder:
Do we need splitting/merging for both Imaging and Epochs?
And what do they mean exactly: e.g. if two Imaging object with one Epoch are merged, does the result have 2 Epochs, or 1 longer Epoch?

alejoe91:
In spikeinterface we have to concepts:
- `append`: makes multi epochs (e.g., imaging with 2 epochs appended to imaging with 3 epochs -> imaging  with 5 epochs)
- `concatenate`: makes a single-epoch imaging, with virtual concatenation of epochs

alessandrofelder:
That nomenclature works for me.
Are there splitting equivalents for e.g. splitting Epochs within an Imaging and splitting Epochs out into separate Imaging objects?

alejoe91:
Yes, we can have a `split_epochs` function that works as follows:

```
def split_epochs(multi_epochs_imaging, epoch_indices: int | list[int]):
    ...
```
It will return an imaging object with 1 or len(epoch_indices) epochs.

alejoe91:
I think we can start adding these functionalities in core. @arielleleon @seanmcculloch @grg2rsr @lauraporta any volunteer?

PR #56, opened 2026-04-09
implement basic splitting

### LINKED ISSUE #45: Utility function to split Imaging objects

alessandrofelder:
Closes #45 

Adds a [utility function `split_epochs`](https://github.com/photon-mosaic/photon-mosaic/issues/46#issuecomment-4073940063) to split out Epochs defined by `epoch_indices` via a proxy Imaging class.

I also considered an alternative implementation strategy where `split_epochs` would create a new Imaging object of the same type as the input Imaging object but that proved hard. The current implementation is also more in keeping with our idea of lazy evaluation via chaining objects together, and only `get_series` actually being final (and this is ~what `spikeinterface` does according to my AI assistant (?))

Question for reviewers: should the proxy class `SelectEpochImaging` be moved to `baseimaging.py` (analogous to `SelectROIs` class being in `baserois.py`)?

### DIFF
--- a/src/photon_mosaic/core/split.py
+++ b/src/photon_mosaic/core/split.py
@@ -14,7 +14,8 @@ def _normalize_epoch_indices(epoch_indices: int | list[int], num_epochs: int) ->
         raise ValueError("epoch_indices must contain at least one index")
 
     for epoch_index in normalized_epoch_indices:
-        if not isinstance(epoch_index, int):
+        # isinstance(True, int) returns True, so we need the second check here
+        if not isinstance(epoch_index, int) or isinstance(epoch_index, bool): 
             raise TypeError("All epoch indices must be ints")
         if not 0 <= epoch_index < num_epochs:
             raise IndexError(f"Epoch index {epoch_index} out of range for imaging with {num_epochs} epochs")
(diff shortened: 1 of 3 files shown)


src/photon_mosaic/core/split.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self._parent = imaging
+        self._kwargs = {
+            "imaging": imaging,
+            "epoch_indices": normalized_epoch_indices,
+        }
+
+
+def split_epochs(imaging: BaseImaging, epoch_indices: int | list[int]) -> SelectEpochImaging:
+    """Return a proxy imaging object with only the requested epochs."""
```
alessandrofelder, line 47: This docstring should probably mention that this does not do a deepcopy.

alejoe91 (approved):
LGTM!

alejoe91:
This looks great @alessandrofelder 

> Question for reviewers: should the proxy class `SelectEpochImaging` be moved to `baseimaging.py` (analogous to `SelectROIs` class being in `baserois.py`)?

I would keep it in a separate file. Potentially we should move out the `selectrois` to a separate file too

src/photon_mosaic/core/split.py
alejoe91: This is a global design, so IMO it would be enough to state that globally in the docs

arielleleon (approved):
This looks great!

PR #60, opened 2026-04-15
Suite2p v1 reg

seanmcculloch:


seanmcculloch:
Suite2p's badframes are used to exclude frames for FOV cropping after motion is detected, and to exclude frames from ROI detection (segmentation). They aren't used for excluding from the reference image during registration

https://github.com/MouseLand/suite2p/blob/e6ed85fc81f4d0660e8f3c783a0eb58a4b8c1b4e/suite2p/registration/register.py#L44
https://github.com/MouseLand/suite2p/blob/e6ed85fc81f4d0660e8f3c783a0eb58a4b8c1b4e/suite2p/detection/detect.py#L22

src/photon_mosaic/preprocessing/baseregistrationsettings.py
```diff
@@ -0,0 +1,272 @@
+from pathlib import Path
+from typing import Optional
+
+from pydantic import ConfigDict, Field
+from pydantic_settings import BaseSettings
+
+
+class BaseRegistrationSettings(BaseSettings):
```
alejoe91: Why do we need this? `Suite2pRegistrationSettings` and `CaImAnRegistrationSettings` could directly inherit from `BaseSettings`.

Or is this a placheholder in case we want to unify some params?

src/photon_mosaic/preprocessing/baseregistrationsettings.py
```diff
@@ -0,0 +1,272 @@
...
+    """Base settings class for all registration methods.
+
+    Holds serializable configuration only. The imaging object (BaseImaging)
+    is a runtime dependency passed separately to the registration class.
+    """
+
+
+class Suite2pRegistrationSettings(BaseRegistrationSettings):
```
alejoe91: I would move this to the `suite2p.py` (and rename it `suite2p_registration`)

src/photon_mosaic/preprocessing/baseregistrationsettings.py
```diff
@@ -0,0 +1,272 @@
...
+    )
+    device: str = Field(
+        default="cpu",
+        description="Torch device for registration: 'cpu', 'cuda', or 'mps'.",
+    )
+
+    model_config = ConfigDict(env_prefix="SUITE2P_REGISTRATION_", case_sensitive=False, env_file=".env")
+
+
+class CaImAnRegistrationSettings(BaseRegistrationSettings):
```
alejoe91: Move to `caiman_registration.py`

README.md
```diff
@@ -1,2 +1,13 @@
...
+
+# Installation
+
+```bash
+git clone https://github.com/photon-mosaic/photon-mosaic.git
+cd photon-mosaic
+pip install -e .
+```
+
+Currently, photon-mosaic parallelization depends on an open PR in SpikeInterface ([#4216](https://github.com/SpikeInterface/spikeinterface/pull/4216)). This is installed automatically when installing from source as shown above.
```
alejoe91, line 13:
```suggestion
```
It's now in main, so we can remove (and install from pypi form the next release!!!)

src/photon_mosaic/preprocessing/suite2p_registration.py
```diff
@@ -0,0 +1,397 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+from photon_mosaic.core import BaseImaging, BaseImagingEpoch
+
+from .basepreprocessor import BasePreprocessor, BasePreprocessorEpoch
+from .baseregistrationsettings import Suite2pRegistrationSettings
+
+
+class Suite2PMotion:
```
alejoe91: Do we need a Suite2p-specific class? We could implement a `core.Motion` class that could handle all the suite2p + caiman arguments/methods and potentially implement the interpolation (with various methods?)

src/photon_mosaic/preprocessing/suite2p_registration.py
```diff
@@ -0,0 +1,397 @@
...
+            planes_to_process = list(range(num_planes))
+        elif isinstance(plane_indices, int):
+            planes_to_process = [plane_indices]
+        elif isinstance(plane_indices, slice):
+            planes_to_process = list(range(*plane_indices.indices(num_planes)))
+        else:
+            planes_to_process = list(plane_indices)
+
+        disps = self.motion.displacements[self.epoch_index]
+        output_planes = []
```
alejoe91: maybe pre-allocating the array would be faster? We know the target shape right?

alejoe91 (changes requested):
Thanks guys for working on this! I left a few comments/suggestions. Let's discuss!

src/photon_mosaic/preprocessing/baseregistrationsettings.py
arielleleon: It was a placeholder to unify some parameters but I would like to talk about this when we meet today

src/photon_mosaic/preprocessing/suite2p_registration.py
arielleleon: That's a good point. Let me see what that would look like

src/photon_mosaic/preprocessing/baseregistrationsettings.py
arielleleon: moved to `suite2p_registration`

src/photon_mosaic/preprocessing/suite2p_registration.py
arielleleon: Yeah - it would save as we increase the number of planes. Resolving

src/photon_mosaic/core/registration.py
```diff
@@ -0,0 +1,88 @@
+from typing import Any, Sequence
+
+import numpy as np
+from numpy.typing import NDArray
+
+from photon_mosaic.core import BaseImaging
+
+
+class Motion:
+    """Container for Suite2P motion correction artifacts."""
```
alejoe91:
```suggestion
    """Container for estimated motion signals and metadata."""
```

src/photon_mosaic/core/registration.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __init__(
+        self,
+        imaging: BaseImaging,
+        displacements: Sequence[NDArray[np.floating[Any]]],
+        refAndMasks: Any,
```
alejoe91:
```suggestion
        reference_masks: Any,
```

src/photon_mosaic/core/registration.py
```diff
@@ -0,0 +1,88 @@
...
+        """Store displacement fields and metadata produced by Suite2P.
+
+        Parameters
+        ----------
+        imaging : BaseImaging
+            Imaging object associated with the computed motion.
+        displacements : Sequence[NDArray]
+            Per-epoch displacement arrays shaped ``(frames, planes, 2)`` or ``(frames, 2)``.
+        refAndMasks : Any
+            Suite2P reference images returned by ``_compute_reference_wrapper``.
```
alejoe91:
```suggestion
            Reference images to perform the registration (e.g., images returned by ``suite2p._compute_reference_wrapper``)
```

src/photon_mosaic/core/registration.py
```diff
@@ -0,0 +1,88 @@
...
+        ops : dict[str, Any]
```
alejoe91: I would remove `ops` from here and make it Suite2P agnostic. 

We have 2 options here:
1. expose the relevant fields of the ops that could be useful to keep track of
2. Implement a `Suite2pMotion` that inherits from this class and adds ops

src/photon_mosaic/core/registration.py
```diff
@@ -0,0 +1,88 @@
...
+            Suite2P block definitions when non-rigid registration is enabled.
+        yranges : Sequence | None, optional
+            Per-epoch, per-plane valid pixel row range ``[epoch][plane] -> (ymin, ymax)``.
+        xranges : Sequence | None, optional
+            Per-epoch, per-plane valid pixel column range ``[epoch][plane] -> (xmin, xmax)``.
+        corrected_badframes : Sequence | None, optional
+            Per-epoch, per-plane boolean bad-frame mask ``[epoch][plane] -> (n_frames,)``.
+            Combines the input ``badframes`` with frames flagged by large registration shifts.
+        registration_outputs : dict | None, optional
+            Additional outputs from the registration pipeline.
```
alejoe91: Remove all `suite2p` references and make docstrings mroe general.

src/photon_mosaic/core/motion.py
```diff
@@ -0,0 +1,88 @@
...
+        nonrigid_offsets : Sequence | None, optional
+            Per-epoch, per-plane non-rigid offsets.
+            Structure: ``[epoch][plane] -> (yoff1, xoff1)`` each ``(n_frames, n_blocks)``.
+        blocks : Sequence | None, optional
+            Suite2P block definitions when non-rigid registration is enabled.
+        yranges : Sequence | None, optional
+            Per-epoch, per-plane valid pixel row range ``[epoch][plane] -> (ymin, ymax)``.
+        xranges : Sequence | None, optional
+            Per-epoch, per-plane valid pixel column range ``[epoch][plane] -> (xmin, xmax)``.
+        corrected_badframes : Sequence | None, optional
```
alejoe91, line 45: Are badframes per plane? I would imagine bad frames to be caused by some external stimuli/stimulations and appear on all planes. What do you think?

src/photon_mosaic/core/motion.py
alejoe91, line 1: Can we rename this file `motion.py`? Conceptually, motion estimation comes before the registration (aka correction/interpolation) step

alejoe91 (changes requested):
Thanks @arielleleon 

Another round of small comments and this is good to go for me!

src/photon_mosaic/core/registration.py
arielleleon: Yikes - let me fix this

src/photon_mosaic/core/registration.py
arielleleon: Done

src/photon_mosaic/core/motion.py
arielleleon: I think you are correct

src/photon_mosaic/core/registration.py
arielleleon: I implemented a `Suite2pMotion` that inherits from the `Motion` class in `core`

src/photon_mosaic/core/motion.py
arielleleon: Yes

src/photon_mosaic/core/motion.py
arielleleon: Okay, since `badframes` exist on the time domain, I made `corrected_badframes` a per-epoch mask on `Motion`. I then changed compute_crop to return the user defined `badframes` and frames whose registration shift exceeded `maxregshift`, then unioned across planes. Does that make sense?

alejoe91 (approved):

PR #63, opened 2026-04-16
Add suite2p segmentation

arielleleon:
- adds suite2p segmentation


arielleleon:
@alejoe91 @lauraporta @alessandrofelder

I have segmentation working and would really appreciate your feedback. This is in a draft state until I add tests. 

I currently have the settings users can access in the `suite2psegmentation` module but I am not sure if this is the best idea. I still haven't gone through CaImAn to see if there are overlapping settings that would flag the use of a 

Heads up before you review: this branch includes changes pulled in from @seanmcculloch branch `suite2p-v1-reg` that I needed to work on this. Those files aren't part of this PR's scope — please focus your review on `/photon-mosaic/preprocessor/sutie2p_segmentation.py`, `/photon-mosaic/core/baserois.py` `/photon-mosaic/core/baseimaging.py`

Once their branch lands on main, I'll rebase this PR so the diff shows only my changes.

arielleleon:
Just realized I may have made an error. I don't think I have allowed for segmentation per epoch. Is that something we would want? How do people run segmentation on volumetric data? I just don't know a lot about this

lauraporta:
Hi @arielleleon, I'll give a look at this PR this week!

lauraporta:
Hi @arielleleon @seanmcculloch, thanks for moving this forward! 😄

A few comments/thoughts while the PR is still in a draft stage to iterate on the current development. 

First thing that I notice is that segmentation is diverging from the Imaging/Epoch classes duality, so I got a bit lost. With Epochs we can ensure that we are using a homogeneous API to use a certain time and plane window, which is currently the case for registration. With the current implementation we are always taking the whole movie it seems to me... Segmenting a 2GB video pushed my swap to ~+20GB. If we are using a pre-computed average image and cellpose-SAM it should be possible to segment a chunk of frames at a time. Not sure we can have full lazyness though for all the segmentation strategies.

I think this architecture discrepancy / lack of lazyness boils down to using the `detection_wrapper` from suite2p. I think we could do as for registration, use the inner functions themselves, and find a way to have back the `get_series()` functionality here as well.

Then, ideally we want to be flexible on suite2p native segmentation backends, cleanly separate sparsery, sourcery, and cellpose methods. That should be also possible though settings.

Regarding the Epochs question, yes I think segmentation should be applied separately across epochs. I think the concatenation bit we discussed can happen with a merging tool. So merge (concatenate) epochs first inside an imaging object, and keep track of the timings, and then split the epochs again.

There's an open broader question on how we keep this architecture flexible on future new backends (e.g. CaImAn).

Regarding the settings I think we might want to include them as metadata of the imaging object in order to track them via the json files/ provenance idea.

Happy to hear more opinions about this and to chat further!

arielleleon:
> Hi @arielleleon @seanmcculloch, thanks for moving this forward! 😄
> 
> A few comments/thoughts while the PR is still in a draft stage to iterate on the current development.
> 
> First thing that I notice is that segmentation is diverging from the Imaging/Epoch classes duality, so I got a bit lost. With Epochs we can ensure that we are using a homogeneous API to use a certain time and plane window, which is currently the case for registration. With the current implementation we are always taking the whole movie it seems to me... Segmenting a 2GB video pushed my swap to ~+20GB. If we are using a pre-computed average image and cellpose-SAM it should be possible to segment a chunk of frames at a time. Not sure we can have full lazyness though for all the segmentation strategies.
> 
> I think this architecture discrepancy / lack of lazyness boils down to using the `detection_wrapper` from suite2p. I think we could do as for registration, use the inner functions themselves, and find a way to have back the `get_series()` functionality here as well.
> 
> Then, ideally we want to be flexible on suite2p native segmentation backends, cleanly separate sparsery, sourcery, and cellpose methods. That should be also possible though settings.
> 
> Regarding the Epochs question, yes I think segmentation should be applied separately across epochs. I think the concatenation bit we discussed can happen with a merging tool. So merge (concatenate) epochs first inside an imaging object, and keep track of the timings, and then split the epochs again.
> 
> There's an open broader question on how we keep this architecture flexible on future new backends (e.g. CaImAn).
> 
> Regarding the settings I think we might want to include them as metadata of the imaging object in order to track them via the json files/ provenance idea.
> 
> Happy to hear more opinions about this and to chat further!

Thanks for looking at this!


Okay - The `detection_wrapper` has been unwrapped and added to segmentation. There is now a private function, `_stream_bin_movie` that pulls frames in batches via the epoch `get_series` API so the raw movie is never resident - only the binned output. The detection step is replaced by direct calls to `sparsedetect.sparsery` / `sourcery.sourcery` / `antomical.select_rois` by way of `_detect_rois_from_mov`. 

I added a `scope="per_epoch"` that returns `Suite2pEpochSegmentations` plus a `scope="all_epochs"` that concatenates internally. We should discuss the approach you mentioned since it requires a change to the imaging API.

I left the algorithm key in the settings. Suite2p only offers these three segmentation algorithms and separating them seems a bit much but I'm happy to discuss this when we meet

src/photon_mosaic/segmentation/suite2p_segmentation.py
alejoe91, line 1: @arielleleon I think this should go to a `segmentation` module, instead of `preprocessing`

arielleleon: Yes - great point, I will refactor now

alejoe91: @arielleleon I think you moved this to extractors, but it should go into a separate `segmentation` module. Extractors should merely be about IO stuff (mainly inputs)

arielleleon: @alejoe91 should be fixed now

lauraporta:
Thanks @arielleleon! Reviewing it today!

lauraporta (changes requested):
Hi @arielleleon, thanks for continuing iterating on this PR! 
In general it looks good to me.
Would it be possible to fill in `test_suite2p_segmentation.py` before merging? Maybe you have some uncommitted changes locally, currently the file is empty in the diff. 

After this is merged, the next step forward is ROI classification! (is_cell)


Edit: also we might want to discuss if `suite2p` should be as well one of the main dependencies from now on, not just in `suite2p-registration` optional dependency 🤔

src/photon_mosaic/segmentation/suite2p_segmentation.py
```diff
@@ -0,0 +1,647 @@
...
+    )
+
+    model_config = SettingsConfigDict(
+        env_file=".env", env_prefix="SUITE2P_SEGMENTATION_", case_sensitive=False, extra="ignore"
+    )
+
+    @field_validator("algorithm")
+    @classmethod
+    def lowercase_algorithm(cls, v: str) -> str:
+        return v.lower()
```
alessandrofelder: Maybe nitpicky: could do with a docstring in case we ever want to auto-generate API reference? (Or mark this as private with an _?)

lauraporta (approved):
LGTM!

src/photon_mosaic/segmentation/suite2p_segmentation.py
```diff
@@ -0,0 +1,647 @@
...
+                "soma_crop": self.soma_crop,
+                "allow_overlap": self.allow_overlap,
+                "nbins": self.nbins,
+                "highpass_time": self.highpass_time,
+            }
+        )
+        return base
+
+
+class Suite2pDetectedRois(BaseRois):
```
alejoe91, line 107: this should be a class method of the `extractors.Suite2pRois.from_stats()` (and this should be re-added to the extractors!)

PR #69, opened 2026-05-05
Overview of `photon-mosaic` high-level design, for developers

### LINKED ISSUE #41: Developer docs: describe mental model
This should include how we think about Imaging, Epochs, ROIs and Analyzer classes and how they related to the video data

alessandrofelder:
I think this is now in advanced draft form, and I think it gives a reasonable high-level overview of the architecture of the project, for newcomer developers.

~~Very much work-in-progress for now! Didn't get as much time on this as I would have liked.~~

~~Sections I still want to add~~

* [x] more detail on how we can lazily nest operations and keep track.
* [x] `photon-mosaic`-native IO to zarr/binary
* [x] BaseROIs
* ~~[ ] How BaseROIS know about Imaging metadata~~
* [x] be refactored/transferred to show in dev docs on our website.

Closes #41


alessandrofelder:
> Maybe we also add tutorial notebooks as sphinx gallery for each of the concepts, in separate issue/PR.

Opened #122 for this

alessandrofelder:
Tagging @jeromelecoq as promised in the last developer meeting :)

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+
+The [`segmentation` module](./api/segmentation.rst) provides functionality that spatially segments `Imaging` objects and returns a `ROIs` object.
+
+#### `Analyzer` and `AnalyzerExtension` objects
+
+`Analyzer` objects combine `Imaging` and `ROIs`. They provide a `compute` function that takes the name of an `AnalyzerExtension` as a string (and possibly more arguments). `AnalyzerExtension`s may depend on each other (e.g. the `df_over_f` extensions depends on the `fluorescence_trace` extension), and calling `compute` on an extension that depends on another extension that is not yet computed will raise an error.
+
+### Reading and writing data
+
+The [`extractors` module](./api/extractors.rst) reads data written by external software (e.g. ScanImage acquisitions or Suite2p ROIs). We rely on the [`roiextractors` package](https://roiextractors.readthedocs.io/en/latest/index.html), which we dynamically wrap, to support reading a wide range of raw input time series. In accordance with the mental model above, we ensure we return `Imaging` objects with appropriate metadata, containing `Epoch`s that are always four-dimensional arrays (time, height, width, planes), lazily.
```
jeromelecoq: From the documentation, I am reading a single time is associated with an image. Is that correct? 

One could take into account the scanning pattern of a given imaging modality. In this model, each pixel gets a timestamp. To do this, one would have to load data with an understanding of how it was acquired and scanned. This is therefore extra-work that most analysis package do not consider. In 2P scanning, the top left pixel is acquired almost at the same time as the previous bottom-right pixel but often they are given a middle time value.

If such a structure is set in place, the Imaging object would store an array of "delta-t", this would present the temporal shift from the average time of all pixel. 

One could then project with a ROI both the pixel and time matrix to extract time AND value for a given ROI. Alternatively, the movie can be realigned in a pre-processing step so that all pixel are interpolated at the same time. 

Doing this right could add value to the field overall as we could report time of each ROI correctly.

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+
+![](_static/single-plane-single-epoch.png)
+![](_static/multi-plane-single-epoch.png)
+
+Single- and multi-plane time series are stored in the same epoch...
+
+![](_static/single-plane-single-epoch.png)
+![](_static/single-plane-single-epoch.png)
+
+... but different channels, and time series that were interrupted (e.g. consecutive days) belong in different `Epoch`s, but the same `Imaging` object (assuming they were taken from the same anatomical location).
```
jeromelecoq: the term Epoch initially confused me. I understand that it refers to a block of data acquired continuously in one region of space. 
This confused me because I was assuming an Epoch exclusively refer to time range not space, so I was expecting a multi-channel recording to be a single Epoch.

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+| Core |  Base class definitions, analyzers, core utility functions |
+| Extractors | Classes that allow extraction of `photon-mosaic` objects from third party tools |
+| Preprocessing | Motion correction and other pre-processing steps (return `Imaging` objects) |
+| Sample Data | Provision of public example data for manual testing and examples |
+| Segmentation | ROI segmentation algorithms (take `Imaging`, return `ROIs`) |
+| Widgets | Interactive IPython widgets for plotting in notebooks |
+
+Note that each submodule has its own `tests/` folder for unit tests.
+
+### The mental model for `Imaging` and `Epoch` objects
```
jeromelecoq, line 66: Diagrams of the classes could be useful. There are 2 views that can be referred to : 
1/ A class structure hierarchy: Who inherits from whom?
2/ How the classes feed into each other, ie how they are dominoed to process data.

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+
+`photon-mosaic` takes an object-oriented approach to represent multi-photon time series. Two types of object are central to this: `Imaging` objects represent a collection of  multi-photon time series that are (approximately) contiguous in space. `Epochs` are contained within `Imaging` objects and represent multi-photon time series that are (exactly) contiguous in space _and_ channel _and_ time. This concept is visualised below:
+
+![](_static/single-plane-single-epoch.png)
+![](_static/multi-plane-single-epoch.png)
+
+Single- and multi-plane time series are stored in the same epoch...
+
+![](_static/single-plane-single-epoch.png)
+![](_static/single-plane-single-epoch.png)
```
lauraporta: I think this is a copy-paste bug: this second pair shows `single-plane-single-epoch.png` twice, but the caption below talks about different channels and interrupted acquisitions. The PR adds `multi-channel-multi-epoch.png` and `multi-aquisition-multi-epoch.png` but never references them — I guess these are the intended images? (The local Sphinx build confirms only 2 of the 4 new PNGs end up in the site.)

docs/source/contributing.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Building documentation locally
+
+The documentation is found in the `docs/source` folder, where the structure mirrors the rendered website.
+
+Dependencies for building the documentation locally can be found at `docs/requirements.txt`.
+To install these, change directory to the `docs` folder in your terminal and type:
+
+```
+pip install -r requirements.txt
```
lauraporta: This command fails as written: `requirements.txt` starts with `-e .[widgets]`, which resolves relative to `docs/` and errors with "does not appear to be a Python project" (reproduced with pip in a fresh venv — I think the instructions came from the pipeline repo, where the layout differs). Running `pip install -r docs/requirements.txt` from the repo root works, so either the instruction could say that, or the requirements line could become `-e ..[widgets]`.

docs/source/contributing.md
```diff
@@ -0,0 +1,66 @@
...
+If you're unsure about any part of the contributing process or have any questions, please
+get in touch through our [Zulip chat](https://neuroinformatics.zulipchat.com/#narrow/channel/500681-photon-mosaic).
+Otherwise, feel free to dive right in and start contributing by
+[creating a development environment](#creating-a-development-environment)
+and [opening a pull request](#pull-requests).
+
+## Creating a development environment
+
+To install ``photon-mosaic`` for development, first the
+[GitHub repository](https://github.com/photon-mosaic/photon-mosaic-pipeline)
```
lauraporta: This links to the `photon-mosaic-pipeline` repo — should be `photon-mosaic`.

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+### Exposing a functional API to users
+
+Outward-facingly, we prefer to expose functions rather than classes to users and libraries depending on `photon-mosaic`. This is because we think this is more elegant and user-friendly.
+
+Pattern
+
+```python
+class DoSomething:
+
+do_something = DoSomething
```
lauraporta: This block renders as invalid Python (the class has no body), which will confuse exactly the newcomer audience it's aimed at — maybe `class DoSomething:` / `    ...` / `do_something = DoSomething`? Also, the alias in `binaryimaging.py` is called `read_binary_folder`, not `read_binary_imaging`.

docs/source/_static/multi-acquisition-multi-epoch.png
lauraporta, line 1: I cannot see the text when I am in dark mode

Image

docs/source/user-guide.md
```diff
@@ -0,0 +1,16 @@
+# Usage
+
+## Installation
+
+We currently support Python versions 3.12, 3.13, and 3.14. There is currently no release on PyPI yet, but you can install `photon-mosaic` from Github.
+We strongly recommend that you use an environment manager (e.g. `conda` from [Miniforge](https://conda-forge.org/download/)). In your terminal, you can then run:
+
+```bash
+conda create -n photon-mosaic
```
lauraporta, line 9: Here `conda` but in the dev guides we suggest `uv` 🤔 I'm fine with either choice but maybe we want to only suggest one

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+### Prerequisites
+
+Initial familiarity with
+- object-oriented software development
+- lazy evaluation of operations on chunked arrays
+- multi-photon times series analysis
+
+### Multi-photon time series analysis
+
+On a conceptual level, analysing multi-photon time series typically means performing a series of image processing steps resulting in a number of regions of interest (ROIs). From each ROI, we then extract its (one-dimensional) temporal signal ("trace"), which can be processed further.
```
lauraporta, line 33:
```suggestion
On a conceptual level, analysing multi-photon time series typically means performing a sequence of **image processing steps** with the goal of identifying regions of interest (ROIs), groups of pixels corresponding to our putative neurons. These steps usually include algorithms performing the following tasks: registration, segmentation and ROI classification. From each ROI, we then extract its (one-dimensional) temporal fluorescence signal ("trace"), a proxy of their firing activity, which can be processed further.
```

lauraporta: I have expanded the description as just after we start to talk about segmentation but the concept was never mentioned above.

lauraporta: In my conceptualisation of the pipeline, signal extraction is part of the "image processing steps". This is also the mental model that I used in the schematics. But in our architecture it comes though the `Analyser` classes which is signal processing. I think this interpretation might be correct as well. In any case I think text and schematics should be consistent, so one of them will need to be aligned.

lauraporta: I wonder if a broad pipeline schematic would be also useful at this point.

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+
+:::{note}
+While we can still _conceptually_ think of analysis as a linear sequence of steps, some algorithms simultaneously segment the ROIs and extract their temporal signal, e.g. by minimising a function that depends on both.
+:::
+
+For image processing, `photon-mosaic` defines its own data structures to represent multi-photon time series (`Imaging` and `Epoch`) and ROIs (classes derived from `BaseROIs`). Signal processing steps are encoded in `AnalyzerExtensions` which are wrapped into an `Analyzer` class.
```
lauraporta: I think this can be a good place in which to include a schematic like the one used in the poster.

docs/source/developer-intro.md
lauraporta, line 66: Great section!

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+
+For example, let's say someone acquires a 2-channel time series of a certain brain region. The acquisition is repeated daily, on the same animal and region, for three days. In this case, each day's time series would be two epochs (one per channel) containing all planes, and all the data would live in the same `Imaging` object, which would contain six epochs (3 days by 2 channels).
+
+Time series of different brain regions should be kept in separate `Imaging` objects. Sometimes, adhering to this mental model requires splitting or merging of input data. `photon-mosaic` provides functionality for this.
+
+Beyond grouping `Epoch`s, `Imaging` objects have the additional responsibility for keeping track of metadata around the operations applied to them. They do this via updating their internal dictionary `self._kwargs`.
+This enables both provenance tracking and lazy evaluation. Only when calling `.get_series()` on an Imaging object while the stored chain of operations be executed and the requested array data be loaded into memory as needed.
```
lauraporta: This sentence says a lot of important concepts that might deserve more space.

I think the `.get_series()` and laziness in general hold a lot of complexity, I wonder if we could expand a little and/or reference to spikeinterface docs. 

We can give a definition of provenance for example, of lazy evaluation and explain why this matters in the context of large image arrays. And the indexing and chunking by time and plane... After all the available software to process multiphoton data (as far as I know) offer limited tooling to interact with arrays in this manner.

It will be cool to expand also on the "russian doll" aspect of the `.get_series()` method.

docs/source/developer-intro.md
lauraporta: However, when actually reading ROIs info pre-computed by suite2p we have our own implementation, we do not rely on roiextractors

docs/source/developer-intro.md
```diff
@@ -0,0 +1,118 @@
...
+do_something = DoSomething
+```
+
+For external collections of classes we might want to wrap, this can be done dynamically, using the Python standard library's `inspect` module.
+Examples of this include:
+
+* [(static) `read_binary_imaging/BinaryFolderImaging` ](https://github.com/photon-mosaic/photon-mosaic/blob/43cfc029f16639b7a9a8258207e5798803c3486f/src/photon_mosaic/core/binaryimaging.py) (line 257)
+
+* [(dynamic) extractor classes for `roiextractors`](https://github.com/photon-mosaic/photon-mosaic/blob/43cfc029f16639b7a9a8258207e5798803c3486f/src/photon_mosaic/extractors/roiextractors.py) (line 55)
+*
```
lauraporta: final dot with nothing

docs/source/contributing.md
```diff
@@ -0,0 +1,66 @@
...
+pip install -e '.[dev]'
+```
+
+Finally, initialise the pre-commit hooks:
+
+```bash
+pre-commit install
+```
+
+## Pull requests
```
lauraporta, line 35: Shall we say to open an issue first to discuss?

lauraporta (changes requested):
Thanks @alessandrofelder! Nice work. A few comments on things to improve or fix

docs/source/user-guide.md
alejoe91: let's suggest uv

docs/source/contributing.md
alejoe91: it should be from the root folder: `pip install -r docs/requirements.txt`

joannercsheppard:
Read through the docs and it reads pretty well to a complete newbie (thanks @alessandrofelder!). A few small notes:
-  I built env using `conda` and it worked fine. Can rebuild with `uv` if you'd like someone to check?
- There are sections that refer to `Epoch` and `Segment` as plural like "`Epoch`s". It looks a little clunky and is readable without the "s".
- The images used for "The mental model for Imaging and Epoch objects" don't seem to explain the concepts super well. I wouldn't be able to work out the relationship between an Epoch and Imaging from the diagrams, or how data segments through Epochs. However, the text does a good job at explaining the relationship between the objects anyway.

docs/source/contributing.md
alessandrofelder: https://github.com/photon-mosaic/photon-mosaic/issues/131

docs/source/developer-intro.md
alessandrofelder: Great point. Have reformulated accordingly.

docs/source/developer-intro.md
alessandrofelder: I agree! Thanks for catching!

docs/source/developer-intro.md
alessandrofelder: > In my conceptualisation of the pipeline, signal extraction is part of the "image processing steps"... But in our architecture it comes though the Analyser classes which is signal processing.

I think I agree with you that it's (mostly) image processing. I don't think we need to restrict Analyzers to be signal-processing-only, so I'll adapt the text accordingly.

docs/source/user-guide.md
alessandrofelder: If we move to users using `uv` we also need to update https://pipeline.photon-mosaic.org/user_guide/installation.html for consistency :thinking:

docs/source/developer-intro.md
alessandrofelder: Great point.

I've split out provenance tracking into its own section (because ROIs object also kind of do this) and expanded on the laziness here, with an example, and mentioning the nesting. I tried not to go too deep (this is a high-level intro) or too broad (concept of laziness is mentioned a prerequisite).

I've made an issue (https://github.com/photon-mosaic/photon-mosaic/issues/136) to add chunking and indexing to the dev docs for when we have a clearer idea and have reached a conclusion on #52 and related things.

alessandrofelder:
Thanks for your reviews everyone! :raised_hands:

>    I built env using conda and it worked fine. Can rebuild with uv if you'd like someone to check?

I don't think this is needed for now, until we settle on the recommended env across the org. (I use `uv` and the CI uses plain `pip` inside `tox` so I think we're covered?)

>    There are sections that refer to Epoch and Segment as plural like "Epochs". It looks a little clunky and is readable without the "s".

Did my best to improve this.

>   The images used for "The mental model for Imaging and Epoch objects" don't seem to explain the concepts super well. I wouldn't be able to work out the relationship between an Epoch and Imaging from the diagrams, or how data segments through Epochs. 

Pushed graphics improvements to #135 

> However, the text does a good job at explaining the relationship between the objects anyway.

Yay!

alessandrofelder:
I think this is ready for another round of review. I've moved some work to follow-on issues in the hope that this is a good enough to merge and be useful for newcomers in the current state :)

alejoe91 (dismissed):
Look nice!!! :)

docs/source/developer-intro.md
```diff
@@ -0,0 +1,138 @@
...
+| Preprocessing | Motion correction and other pre-processing steps (return `Imaging` objects) |
+| Sample Data | Provision of public example data for manual testing and examples |
+| Segmentation | ROI segmentation algorithms (take `Imaging`, return `ROIs`) |
+| Widgets | Interactive IPython widgets for plotting in notebooks |
+
+Note that each submodule has its own `tests/` folder for unit tests.
+
+### The mental model for `Imaging` and `Epoch` objects
+
+`photon-mosaic` takes an object-oriented approach to represent multi-photon time series. Two types of object are central to this: `Imaging` objects represent a collection of  multi-photon time series that are (approximately) contiguous in space. One or several `Epoch` objects are contained within an `Imaging` and represent multi-photon time series that are (exactly) contiguous in space _and_ channel _and_ time. This concept is visualised below:
```
alessandrofelder, line 68: Channels go in separate imaging objects!

docs/source/developer-intro.md
alessandrofelder: This is a great point, and our current design doesn't preclude building in this functionality (I imagine it would probably be part of the preprocessing module). Let's move the discussion to #140

docs/source/developer-intro.md
alessandrofelder: As discussed a few weeks ago, I think we want different channels to live in entirely separate Imaging objects.

docs/source/developer-intro.md
alessandrofelder: Moved to #153 to keep PRs small and transient.

docs/source/contributing.md
alessandrofelder: I think this suggestion has remained valid after discussion in #131 so I suggest we keep this in?

alessandrofelder:
Thanks again for in-depth reviews @lauraporta @joannercsheppard @alejoe91 @jeromelecoq :star_struck: 

I've now corrected the docs to account for the fact that different channels should live in different imaging objects, and I think we have issues open for any concerns I have not yet addressed here (will keep working on the docs side of things).

I also think after the discussion on #131 we can keep the relevant contributing section I added here and close that issue too!

Let me know if you feel there's is anything that should still be addressed as part of this PR.

PR #83, opened 2026-05-29
Add Suite2pImaging extractor for registered binary movies (#77)

### LINKED ISSUE #77: Add Suite2pImaging extractor

lauraporta:
Load a pre-computed suite2p run into photon-mosaic — single-plane atom, compositional.

- **`Suite2pImaging`** — one plane as a lazy, memmapped `BinaryImaging` (`is_registered=True`).
- **`concatenate_planes`** — stack planes into a lazy `(T, H, W, n_planes)` volume (a `core` primitive).
- **`read_suite2p_full(folder)`** — load + stitch a whole run (one object per channel).
- **`into_epochs()`** — one epoch per source file from `ops['frames_per_file']`.
- Notebook `examples/suite2p_extractor.ipynb`: per-plane+stitch, whole-folder, per-file epochs, and per-plane ROI pairing (`read_suite2p_rois(plane).register_imaging(plane)`).

ROIs pair **per plane** only; a volumetric ROI object is deferred to #108. Extractor + core tests pass.


src/photon_mosaic/core/concatenate.py
```diff
@@ -0,0 +1,139 @@
+"""Concatenate imaging objects along the plane axis.
```
alejoe91, line 1: @lauraporta I would rename this `aggregate_planes` or `stack_planes`. In SpikeItnerface (and maybe here too!) we use `concatenate_*` for concatenation in time. What do you think?

src/photon_mosaic/core/concatenate.py
```diff
@@ -0,0 +1,139 @@
...
+        start_frame: int,
+        end_frame: int,
+        plane_indices: slice | np.ndarray | None = None,
+    ) -> np.ndarray:
+        parts = [pe.get_series(start_frame, end_frame, None) for pe in self._parent_epochs]
+        full = np.concatenate(parts, axis=-1)
+        if plane_indices is None:
+            return full
+        if isinstance(plane_indices, slice):
+            return full[..., plane_indices]
```
alejoe91, line 53: Wouln't it be better to extract only the requested planes if `plane_indices` is not None?

alejoe91 (changes requested):
@lauraporta a couple of comments.

Why not making 2 PRs? One for split_epochs/stack_planes and one for suite2p stuff? They seem very orthogonal

src/photon_mosaic/core/concatenate.py
lauraporta: I like `stack_planes`!

src/photon_mosaic/core/concatenate.py
lauraporta: Ah definitely yes, good catch!

lauraporta:
Going to split this PR in two as suggested.

lauraporta:
Content moved to #118 and #119. Closing.

Issue #20, opened 2026-01-23
Testing data


lauraporta:
So far the tests use a randomly generated `numpy` array. This addresses fundamental checks in terms of shape and basic operations. in the future we might like to include also some more realistic synthetic movies (with simulated fluorescent cells, with different overlaps) and open data to be fetched by AWS or Gin

alejoe91:
Agreed. This is mainly for testing, but we could have some much more refined simulators incorporated. In `SpikeInterface`, we have a full `generators` module

PR #90, opened 2026-06-18
Example suite2p data

alessandrofelder:
This PR:
* Provides a python API to get some official suite2p data to play with
* It caches the data with `pooch`
* Provides an example notebook running `suite2p` registration and segmentation under the hood.

### Context

This is a first go at

* providing us with some real-life data to play with, that we can re-use and evolve in a version-controlled way (part of #20 )
* providing users with example notebooks to showcase `photon-mosaic`s functionality


src/photon_mosaic/sample_data/suite2p_sample_data.py
```diff
@@ -0,0 +1,73 @@
...
+import pooch
+
+from photon_mosaic.extractors import read_tiff_imaging
+
+_FOLDER_ID = "0B649boZqpYG1R3ota25jdUthSzQ"
+_RESOURCE_KEY = "0-wSoqFv5rnE6TERPcJHwQtQ"
+_CACHE_DIR = Path(pooch.os_cache("photon-mosaic")) / "suite2p_sample_data"
+
+
+def suite2p_google_drive(num_files: int = 5) -> list:
```
alejoe91:
```suggestion
def download_suite2p_google_drive(num_files: int = 5) -> list:
```
or `get_suite2p_google_drive`?

src/photon_mosaic/sample_data/suite2p_sample_data.py
```diff
@@ -0,0 +1,73 @@
...
+    for index, (file_id, resource_key) in enumerate(file_infos, start=1):
+        file_name = f"file_00002_{index:05d}.tif"
+        local_path = pooch.retrieve(
+            url=f"https://drive.usercontent.google.com/download?id={file_id}&confirm=t&resourcekey={resource_key}",
+            known_hash=None,
+            fname=file_name,
+            path=_CACHE_DIR,
+            progressbar=True,
+        )
+        local_paths.append(local_path)
```
alejoe91, line 71: I think it would be cool to cache (maybe option`cache=True`) the downloaded tiffs in `~/.photon-mosaic/sample_data/suite2p/`

alejoe91 (changes requested):
A couple of suggestions @alessandrofelder !

src/photon_mosaic/sample_data/suite2p_sample_data.py
alessandrofelder: `pooch` caches by default so I don't think we need to add an extra option. But I have changed the `_CACHE_DIR` do your suggestion.

alejoe91 (approved):

PR #92, opened 2026-06-25
Multi-field (mesoscope) imaging + ROI analysis [WIP/draft]

grg2rsr:
> **Draft / work-in-progress.** This is a checkpoint to share direction and get
> early feedback. Public APIs, names, and on-disk layout are provisional and
> will change. Not ready to merge.

> **⚠️ Requires the ScanImage multi-ROI support from `roiextractors` PR
> [catalystneuro/roiextractors#597](https://github.com/catalystneuro/roiextractors/pull/597).**
> `ScanImageMultiROIImagingExtractor` and the per-ROI geometry/timing this PR reads
> off it are introduced there and are not on any `roiextractors` release yet. You
> must install it before this code will import or run, e.g.:
>
> ```bash
> pip install "roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@refs/pull/597/head"
> ```
>
> (or an editable local checkout of that branch). This is a hard build/runtime
> requirement, not just for the tests.

## Motivation

Mesoscope / ScanImage multi-ROI acquisitions capture several **independent
fields of view** in one recording — each with its own pixel grid and physical
position in the sample. These cannot be honestly stacked into a single
`BaseImaging` (they share neither a pixel grid nor a regular depth axis).

This PR introduces a **bag-of-fields** model: each field stays its own
`BaseImaging`, and lightweight container classes add cohesion (shared identity,
selection, geometry) on top.

## Scope: planar first, volumetric-ready

The current goal is **planar** multi-field data. The design is deliberately
shaped so that extending to **volumetric** fields (per-field multi-plane
z-stacks) later is a natural next step rather than a redesign, by keeping two
levels of multiplicity separate:

- **within a field** — the `planes` axis of one `BaseImaging` (where a future
  z-stack would live);
- **across fields** — this container (free pixel shape, free lateral position).

Accordingly, `FieldGeometry` carries the field's **lateral** placement and
identity (`center_xy`, `size_xy`, `affine`, `name`, `uuid`) in scanner-angle
units, with an optional `microns_per_scanner_unit` scale powering derived `*_um`
views; depth is left to a field's own plane axis. Full volumetric mROI is **not**
implemented yet and will first need support in `roiextractors`; this PR just lays
the foundation on the photon-mosaic side.

### `core/multifield.py` — `MultiFieldImaging` + `FieldGeometry`

A container holding one independent `BaseImaging` per field alongside a parallel
`FieldGeometry` table. It offers `len`/iteration/indexing, `iter_fields()` →
`(geometry, imaging)`, selection by `index`/`name`/`uuid`, a `geometry_table()`
(pandas if available), and basic validation (field type, length match, a warning
on heterogeneous sampling frequency).

### `extractors/scanimage_multifield.py` — `multifield_from_scanimage()`

A factory that builds a `MultiFieldImaging` from one or more ScanImage multi-ROI
TIFFs. Each scan field is loaded as a standalone `BaseImaging` through
photon-mosaic's dynamically generated ScanImage loader; its geometry (name,
uuid, center/size, affine, microns-per-scanner-unit from `SI.objectiveResolution`)
is read off the underlying `roiextractors` extractor (requires PR
[catalystneuro/roiextractors#597](https://github.com/catalystneuro/roiextractors/pull/597)
— see the requirement note above). Passing all files of a multi-file recording
ties their samples together so each field spans the whole recording.

**Temporal handling** (`_attach_field_timestamps`): the extractor resolves a
base per-frame timeline (a custom override, else the file's native
`frameTimestamps_sec`, else nominal `arange / fs`) and this is mirrored onto each
field's own independent time vector. Two modes:

- `"per_frame"` (default) — every field shares the same per-frame timeline.
- `"per_field"` — adds the within-frame scan offset
  (`line_period × roi_row_start`), modelling that fields are scanned
  sequentially top-to-bottom within a frame.

An optional custom full-frame timestamp array (or `.npy` path) can override the
native timeline and is shared across fields (then offset per field in
`"per_field"` mode).

### `core/roianalyzer.py` — `MultiFieldRoiAnalyzer` *(design up for discussion)*

The analyzer-side glue class, mirroring `MultiFieldImaging`. It pairs one
**independent** `RoiAnalyzer` per field with the geometry table and re-exposes
the same cohesion surface (`len`/iter/indexing, `iter_fields()`,
`get_analyzer(index|name|uuid)`). Its `compute()` is a thin **fan-out** that
delegates to each field's `RoiAnalyzer.compute()` — results stay on their own
field's analyzer. `create_multifield_roi_analyzer()` now returns this container
instead of a bare list. Field-selection logic is factored into a shared
`_resolve_field_selector` helper used by both containers.

**This piece is the most open part of the PR and is explicitly up for
discussion** — whether a dedicated container is the right abstraction at all, how
per-field analyzers should be grouped, and how cross-field results should
eventually be combined (see below). Feedback very welcome here.

## Tests

Unit tests cover the container mechanics, geometry table, validation, and the
`compute()` fan-out, plus the ScanImage factory. Selector-resolution **semantics**
(case-insensitivity, ambiguity, unknown/empty selectors) are intentionally **not**
re-tested here — they're owned and tested upstream in `roiextractors`; we only
verify photon-mosaic's wiring on top.

## Deliberately deferred (not in this PR)

- **Cross-field combination** of extension outputs — `compute()` is fan-out only.
  Sketch: an on-demand `get_combined_extension_data(name)` on the container + a
  `combine_field_data` hook on `AnalyzerExtension` (combine policy lives on the
  extension), with results keyed by a `(field, roi_id)` MultiIndex. Part of the
  "up for discussion" above.
- **Save/load round-trip** for `MultiFieldRoiAnalyzer` — no `loading.py` dispatch
  branch yet; per-field analyzers are written to `field_<i>/` subfolders that are
  individually loadable via `load_roi_analyzer`.
- **Volumetric mROI** — planar-only for now; needs `roiextractors` support first.

### DIFF
--- a/src/photon_mosaic/core/__init__.py
+++ b/src/photon_mosaic/core/__init__.py
@@ -3,15 +3,14 @@ from .baserois import BaseRois
 from .binaryimaging import BinaryImaging, read_binary
 from .binaryrois import BinaryFolderRois, BinaryRois
 from .generators import generate_imaging_with_rois, generate_random_imaging, generate_rois
+from .geometry import geometry_table
 from .loading import load
 from .motion import Motion
-from .multifield import FieldGeometry, MultiFieldImaging
+from .multifield import fields_containing_point, get_field
 from .numpyimaging import NumpyImaging
 from .roianalyzer import (
     AnalyzerExtension,
-    MultiFieldRoiAnalyzer,
     RoiAnalyzer,
-    create_multifield_roi_analyzer,
     create_roi_analyzer,
     load_roi_analyzer,
     register_result_extension,
--- a/src/photon_mosaic/core/__init__.py
+++ b/src/photon_mosaic/core/__init__.py
@@ -5,10 +5,13 @@ from .binaryrois import BinaryFolderRois, BinaryRois
 from .generators import generate_imaging_with_rois, generate_random_imaging, generate_rois
 from .loading import load
 from .motion import Motion
+from .multifield import FieldGeometry, MultiFieldImaging
 from .numpyimaging import NumpyImaging
 from .roianalyzer import (
     AnalyzerExtension,
+    MultiFieldRoiAnalyzer,
     RoiAnalyzer,
+    create_multifield_roi_analyzer,
     create_roi_analyzer,
     load_roi_analyzer,
     register_result_extension,
--- a/src/photon_mosaic/core/baseimaging.py
+++ b/src/photon_mosaic/core/baseimaging.py
@@ -29,6 +29,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
         self._sampling_frequency = float(sampling_frequency)
         self._shape = tuple(shape)  # Image is intended as a volume (H, W, planes)
         self._average_image = None
+        self._geometry: dict | None = None
 
     def _repr_header(self, display_name=True):
         """Generate text representation of the BaseImaging object."""
@@ -63,7 +64,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
 
         # Format shape string based on whether data is volumetric or not
         shape_repr = f"{shape[0]} rows x {shape[1]} columns "
-        return (
+        header = (
             f"{name}:\n"
             f"{sampling_frequency_repr} - "
             f"{self.get_num_epochs()} epochs - "
@@ -72,6 +73,12 @@ class BaseImaging(BaseExtractor, TimeSeries):
             f"{dtype} dtype - "
             f"{memory_repr}"
         )
+        # Show placement only for fields that carry geometry (multi-field acquisitions).
+        if self._geometry is not None:
+            field_name = self._geometry.get("name")
+            center_xy = self._geometry.get("center_xy")
+            header += f"\n  geometry: name={field_name!r} center_xy={center_xy}"
+        return header
 
     def __repr__(self):
         return self._repr_header()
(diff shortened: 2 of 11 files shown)


grg2rsr:
Reworked per the review feedback — pushed as `493628c` and `dc51fb2`.

**Geometry now lives on `BaseImaging`, not in wrapper objects.** A field is a `BaseImaging` carrying its own placement, so a mesoscope recording is just a `list[BaseImaging]`; the `MultiFieldImaging`/`FieldGeometry`/`MultiFieldRoiAnalyzer` classes are gone (−273 lines).

- `BaseImaging.geometry` — optional dict, backed by a SpikeInterface annotation so it round-trips through `save`/`load`. Contents are JSON-native (`affine` as a 3×3 list of lists).
- `multifield_from_scanimage(...)` now returns `list[BaseImaging]` with geometry set per field.
- Helpers over a field sequence: `geometry_table()` (new `core/geometry.py`, where validation will also go) and `get_field()` + a `fields_containing_point()` stub (`core/multifield.py`).
- Analyzer side is plain Python now — build per-field analyzers with a comprehension; geometry is reachable via `analyzer.imaging.geometry`.

**Persistence:** zarr save/load carries geometry with no extra wiring (verified). Binary-folder doesn't yet (annotations dropped in `_save`/reload) — small follow-up, plus a one-line `include_annotations=True` for the analyzer's zarr path.

**Unchanged:** still requires roiextractors#597; cross-field combination and volumetric mROI still deferred.

PR #93, opened 2026-07-01
naive implementation of plane selector imaging

alessandrofelder:
Adds an imaging object that exposes only certain plane ids from its parent `Imaging`

As I am working towards a flexible function requesting real-life data with a customisable number of planes, regions, and channels (#20), I found this to be useful.

Beyond normal code review, I would be grateful for opinions specifically around whether the `select_planes` function should fulfill

```python
    selected = select_planes(imaging, [2, 0])

    np.testing.assert_allclose(
        selected.get_series(plane_ids=[0]),
        imaging.get_series(plane_ids=[0]), # plane ids persist across selection
    )

```

(the above is the current behaviour) or alternatively we could renumber the planes in the selector internally
```python

    selected = select_planes(imaging, [2, 0])

    np.testing.assert_allclose(
        selected.get_series(plane_ids=[0]),
        imaging.get_series(plane_ids=[2]), # the zeroth selected plane corresponds to id 2 in its parent
    )

```

The added `test_select_planes_allows_subselection_within_selected_proxy` is maybe the easiest place to understand this behaviour.


alessandrofelder:
As discussed in today's developer meeting, `plane_ids` are IDs not indices, so the current behaviour is consistent.

alejoe91 (approved):
LGTM!

Issue #91, opened 2026-06-18
Improve `suite2p` sample data API


alessandrofelder:
Follow-up issue to #90 

- [ ] API should expose `n_frames`, not `n_files`
- [ ] API should concatenate files into a single epoch under the hood
- [ ] Can we make this more memory efficient? May need to write to file or make a better generic TIFF extractor
- [ ] A registry with the SHAs of each file to check download is valid.

alessandrofelder:
- [ ] example notebook should include fluorescence and DF/F analysis/plotting once we have #85 merged (and maybe #98 addressed)

Issue #55, opened 2026-04-09
Deinterleaving functionality


alessandrofelder:
> Not volunteering yet but adding some context in case it helps anyone understanding the scope better and pick up this issue.
> 
> Depending on the kind of microscope in use, it can be possible that multiple areas of the cortex are imaged in a unique section. One case that I have seen is of four different areas in the mouse visual cortex which frame acquisitions were interleaved as if they were a subsequent temporal sequence. (Or at least this is how the `roiextractor` package has been reading the tiff files!) For instance: 
> frame1-> region 1, frame2-> region 2, frame3-> region 3, frame4-> region 4, 
> frame5-> region 1, frame6-> region 2, frame7-> region 3, frame8-> region 4, 
> 
> This mix of the temporal and spatial dimensions breaks one assumption we have verbally agreed on that one `BaseImaging` object should contain data that is continuous in cortical space (e.g. one cortical column, so planes do fit this view).  (`Epoch`s for instance should be contiguous in space and time.)
> 
> An example dataset to play with is mentioned in issue #10. [This gist](https://gist.github.com/lauraporta/41ceb12ab768aa329f4f283db0245a57) can be adapted to download the data. 

 _Originally posted by @lauraporta in [#45](https://github.com/photon-mosaic/photon-mosaic/issues/45#issuecomment-4084842440)_

Splitting this out into a separate issue to simplify discussion. 

The `deinterleave` function should split the 1-epoch Imaging object returned by roiextractors into 4 Imaging objects (because Imaging objects should be continous in space)

alessandrofelder:
I have played with the dataset in question, obtained via:
```python
import pooch
bucket = "aind-open-data"
file_key = "multiplane-ophys_826616_2026-01-02_11-58-49/pophys/1483779513_local_z_stack2.tiff"
url = f"https://{bucket}.s3.amazonaws.com/{file_key}"
local_path = pooch.retrieve(url=url, known_hash=9130959574bce0eff383112d03b0a1b5a9e3ccf0a900fa5658c5919e5c16b177, progressbar=True)
```
and tried to interpret the metadata with Claude, and did some visualisation. I document my findings below to minimise duplication of efforts (hopefully 😁 )

While the metadata seems to tell us there are 4 ROIs and possibly 2 channels, it didn't look like that to me when visualising (note: I loaded the data with `tifffile.imread` rather than `roiextractors` functionality):
[...]

PR #95, opened 2026-07-01
AIND sample data

### LINKED ISSUE #10: link to local stack
We need a test stack to play with volumetric processing

alessandrofelder:
Another step towards finishing #20, follow on from Plane Selector PR #93

Adds a function where you can get any subset of 
- one or two channels (`"channel_1, channel_2")
- one to four regions (0,1,2,3)
- one to 81 planes

And a small related explanatory example (just plotting) - if anyone has suggestions around making the plots more interesting, feel free to comment!

I would argue this closes #10, and is related to https://github.com/photon-mosaic/photon-mosaic/issues/55#issuecomment-4224473102
What's missing from #20 is #91.

Something for @arielleleon to check (that I still don't understand - sorry :sweat_smile:):
* Are the 81 planes within each zstack contiguous in space (and therefore a "true" imaging object)?


src/photon_mosaic/sample_data/aind_sample_data.py
```diff
@@ -0,0 +1,69 @@
...
+        Default is the middle plane.
+
+    Returns
+    -------
+    list[BaseImaging]
+        List of plane-selected imaging objects, one per
+        ``(channel, region)`` combination. The region index changes first.
+    """
+    bucket_name = "aind-open-data"
+    imagings = []
```
alejoe91: I would return this as a dict with tuples as keys: `(channel, region)`

@alessandrofelder what do you think?

alessandrofelder: I think that's a nice idea!

alejoe91 (approved):

alejoe91 (approved):

arielleleon:
> Another step towards finishing #20, follow on from Plane Selector PR #93
> 
> Adds a function where you can get any subset of
> 
> * one or two channels (`"channel_1, channel_2")
> * one to four regions (0,1,2,3)
> * one to 81 planes
> 
> And a small related explanatory example (just plotting) - if anyone has suggestions around making the plots more interesting, feel free to comment!
> 
> I would argue this closes #10, and is related to [#55 (comment)](https://github.com/photon-mosaic/photon-mosaic/issues/55#issuecomment-4224473102) What's missing from #20 is #91.
> 
> Something for @arielleleon to check (that I still don't understand - sorry 😅):
> 
> * Are the 81 planes within each zstack contiguous in space (and therefore a "true" imaging object)?

@alessandrofelder - Not always. That would need to be determined from the metadata of the input data format. For example, in #55, each local z stack contains two planes interleaved where a z stack is in the same FOV (x,y) but at different z locations (e.g. z1=60um and z2=500um). These image stacks would need to be deinterleaved and saved out as separate imaging objects. I don't know whether other labs create deinterleaved stacks that are spatially contiguous which is why I am being careful about checking the metadata.

alessandrofelder:
Thanks @arielleleon - I think what's confusing me is: the `session.json` file is at the level above the actual files, so I assume it pertains to all the files in `multiplane-ophys_826616_2026-01-02_11-58-49`. From the JSON file I do seem to get 4 ROIs (i.e. 4 different values for `scanimage_roi_index` in the metadata), but also there are 4 tiff files ("pophys/ 1483779513_local_z_stack{0-3}.tiff" each with two channels, which my code here splits by specifying `channel=` in `read_scan_imaging`) so my default assumption is that each TIFF channel is a spatially contiguous ROI for this specific example dataset (I appreciate this might be different in other datasets). Would you agree?

arielleleon:
> Thanks @arielleleon - I think what's confusing me is: the `session.json` file is at the level above the actual files, so I assume it pertains to all the files in `multiplane-ophys_826616_2026-01-02_11-58-49`. From the JSON file I do seem to get 4 ROIs (i.e. 4 different values for `scanimage_roi_index` in the metadata), but also there are 4 tiff files ("pophys/ 1483779513_local_z_stack{0-3}.tiff" each with two channels, which my code here splits by specifying `channel=` in `read_scan_imaging`) so my default assumption is that each TIFF channel is a spatially contiguous ROI for this specific example dataset (I appreciate this might be different in other datasets). Would you agree?

This is easy to confuse. The session.json has a `scanimage_roi_index` which indicates which lateral FOV the plane was imaged in. There is also a `coupled_fov_index` which indicates which scan planes are coupled. 

In this case, there are 4 lateral positions (ROIs) with two planes (FOVs). We call it a 4 (ROIs) x 2 (FOV's - imaging planes)) imaging configuration. We can have many combinations (1 x 8, 2 x 4, etc.)

For this session see the table below:

| `scanimage_roi_index`     |     file            |       lateral (x,y)      |     imaging depths      |      `coupled_fov_index`    |
|-------|-----|------------------|-----------------|------------------|
|    0                  |   `local_z_stack0.tiff`              |       (-6.67, -7.21)     |   -126, -244           |      0          |     
|    1                  |   `local_z_stack1.tiff`              |      (-2.37, -4.12)     |   -144, -266         |      1          |     
|    2                  |   `local_z_stack2.tiff`              |      (-10.83, −1.59)     |  -112, -232          |      2          |     
|    3                 |   `local_z_stack3.tiff`              |      (-7.16, 1.71)     |   -138,-250         |      3          |  

Both channels image the same tissue (same lateral location, same depths) meaning that this is a signal axis, not a spatial one. Two depths are the two spatially distinct volumes per ROI. 

In the TIFF file, **within a given channel**, the planes are interleaved into every other channel:

|page  | 0    | 1    |    2    |    3  |
|-------|-----|------------------|-----------------|------------------|
image  | depth  -126   |     depth  -244   |   depth -126   |   depth  -244  |

It would be good to have a deinterleave function that is capable of splitting up multiplane data like this into distinct imaging objects. We should talk about how this lands in photon mosaic and how to make it metadata-driven (topology decision should happen before the object is built).

*note*: don't use the `fov_coordinate_ml`/`ap` in the session.json. It's hardcoded

To take it a step further, I think we could have a generic `deinterleave` function that can take in a numpy array and split data like these with `k` stride (TIFF layout in example above is `k=2`) and offset:

```python
import numpy as np


def deinterleave(frames: np.ndarray, n_streams: int) -> list[np.ndarray]:
    """Split a round-robin interleaved stack into per-stream stacks.

    Assumes acquisition order stream_0, stream_1, ..., stream_{N-1},
    stream_0, ... so that stream k is frames[k::n_streams].

    Parameters
    ----------
    frames : np.ndarray
        Interleaved data, shape (n_frames, H, W). n_frames need not be an
        exact multiple of n_streams; trailing partial cycles are allowed
        (earlier streams simply get one extra frame).
    n_streams : int
        Number of interleaved streams (regions / planes).

    Returns
    -------
    list[np.ndarray]
        n_streams arrays, stream k = frames[k::n_streams].
    """
    if n_streams < 1:
        raise ValueError(f"n_streams must be >= 1, got {n_streams}")
    if frames.shape[0] % n_streams != 0:
        import warnings
        warnings.warn(
            f"{frames.shape[0]} frames is not a multiple of "
            f"{n_streams}; streams will have unequal lengths."
        )
    return [frames[k::n_streams] for k in range(n_streams)]
``` 



Did this answer your question?

alessandrofelder:
Thanks so much @arielleleon - sorry for being slow - I think I understand your explanation and am finally with you.

I think my confusion was arising from my interpretation of the word "channel". I assumed channels meant measuring the emission of light of two wavelengths from the same place but actually it is (possibly the same wavelength of light) measured in two places!

I want to double-check the periodicity of the data though:

Once I separate the channels within local stack 0, and plot the mean intensity over time, I get a very strong periodic signal. Is this expected (e.g. because the experiment involved the delivery of a periodic stimulus of some sort?) 

image

alessandrofelder:
If the periodicity is related to time and not to different planes (as I now think is the case), then there is a bug/misinterpretation of how `roiextractors` reads this data:

```python
imaging = read_scan_image_imaging(
    file_path=local_path,
    channel_name=f"Channel {channel}",
)
print(imaging.num_planes)
```
gives 81 for each channel.

arielleleon:
@alessandrofelder - something is going on with the extractor. I am looking at it now

arielleleon:
> If the periodicity is related to time and not to different planes (as I now think is the case), then there is a bug/misinterpretation of how `roiextractors` reads this data:
> 
> ```python
> imaging = read_scan_image_imaging(
>     file_path=local_path,
>     channel_name=f"Channel {channel}",
> )
> print(imaging.num_planes)
> ```
> 
> gives 81 for each channel.

Okay, so I dug into this and found that the volumetric, interleaved data can be read with `read_multi_tiff_multi_page`. I also realized some other things along the way. 

Here is how you load the data:

```python
image = pme.read_multi_tiff_multi_page(
                                       file_paths=[image_fp],
                                       sampling_frequency=10.63,
                                       dimension_order="ZCT",
                                       num_planes=2
)
```

**The problem** is this:

```python
# All three read both planes and discard one. Pick on ergonomics, not I/O:
plane_0 = image.get_series(plane_ids=[0])           # (n, H, W, 1), whole recording
chunk   = image.get_series(0, 10, plane_ids=[0])    # (10, H, W, 1), first 10 volumes
deep    = select_planes(image, plane_ids=[0])       # reusable handle, same cost per call
```

Both planes are always read in which I don't think we necessarily want. I'm still pondering this but wanted to give you a way to read the data

alessandrofelder:
Thanks a lot for your investigation @arielleleon . I've refactored the files in this PR to reflect my understanding, and to use your suggestion of how to read in the data.

I think the performance is a separate issue to the provision on real-life sample data?

Can I double-check what the explanation for the 81-periodicity is?

arielleleon:
This file is a local z-stack, not a functional recording. The microscope starts at the top depth and images 81 planes down into the tissue, then jumps back to the top and repeats, 20 times. So the 1620 "timepoints" you're plotting aren't time, they're 81 depths, repeated 20×. You are seeing a depth profile where the mean intensity falls off as you go deeper, being re-traced every 81 samples. 

Quick way to confirm: trace.reshape(20, 81) and overlay the rows — all 20 sweeps sit on top of each other.

Separately, I think we're reading the file wrong, which is worth fixing before this merges. It's 3240 pages = 81 slices × 2 channels × 20 volumes, with channel varying fastest. Our current call leaves num_channels at its default of 1, so the two channels come through looking like two planes:
▎
```python
image = pme.read_multi_tiff_multi_page(
    file_paths=[fp],
    sampling_frequency=10.63 / 81,   # 0.1312 Hz volume rate
    dimension_order="CZT",           # channel fastest, then z, then volume
    num_channels=2,
    channel_name="0",                # "1" for red
    num_planes=81,
)
# -> shape (512, 512, 81), 20 samples
```

Let me know if this doesn't make sense

alessandrofelder:
> . The microscope starts at the top depth and images 81 planes down into the tissue, then jumps back to the top and repeats, 20 times. 

This makes total sense and is consistent with my observations - thank you! I think it unfortunately also means that this is probably not appropriate sample data, so I'll look for some different sample data that contains a true recording.

I'll keep this open until we write up an issue around what exactly is wrong with the extractor (I'm not sure I entirely understand that part).

Issue #36, opened 2026-03-02
Look up OMEzarr for storing metadata (and data)


alejoe91:

alessandrofelder:
OME zarr specs at https://ngff.openmicroscopy.org/specifications/0.5/index.html

alessandrofelder:
Outcome of a short investigation and a quick discussion with @K-Meech (thank you!)
* OME-zarr specs would definitely support metadata like time, channels and various spatial axes so would definitely be appropriate
* The software ecosystem continues to be somewhat fragmented, but has a fair amount of community around it
* A significant amount of the metadata is geared towards multiscale data which we don't need for our applications currently (each timepoint easily fits into memory, and I don't see a use case for downsampling in time)

So I think supporting OME-Zarr would be nice for the sake of adhering to community standards, but I don't think it's a priority right now, and it would be easy to add on later if we support `zarr`.

`ome-zarr-models` would be a good library to write the metadata into the `zarr.json`

Issue #35, opened 2026-03-02
Initial DF/F processor


alessandrofelder:
probably per-plane wrapper around suite2p first

alessandrofelder:
Linking to what looks like a [great resource for DF/F](https://millerbrainobservatory.github.io/mbo_utilities/dff.html)  (courtesy of @sumiya-kuroda ) so it doesn't get lost in the depths of slack :)

Issue #33, opened 2026-02-16
Look up for why we open/close memmap in SpikeInterface


alejoe91:
Here: https://github.com/photon-mosaic/photon-mosaic/blob/dev/src/photon_mosaic/core/binaryimaging.py#L141-L188

alejoe91:
This is the issue that explains it:
https://github.com/NeuralEnsemble/python-neo/pull/1265

Issue #32, opened 2026-02-16
Add `write_zarr` to SpikeInterface chunkable branch


alejoe91:
We have machinery in spikeinterface to write to zarr, but it hasn't been ported to the `Chunkable` object yet.

We should make it recording agnostic and wrap it here

PR #101, opened 2026-07-14
feat(core): add DeconvolutionExtension for OASIS-based spike inference

j-friedrich:
## Summary

Adds `DeconvolutionExtension`, a new ROI analyzer extension that infers denoised and deconvolved neural activity from dF/F traces using OASIS (`oasis-deconv`), plus more realistic synthetic-data generation to validate it against known ground truth.

### New: `DeconvolutionExtension`
- Wraps `oasis.functions.deconvolve`; depends on `df_over_f`
- `decay_time`/`rise_time` control calcium kinetics: known values, AR(1) auto-estimate (default, `rise_time=0`), or AR(2) auto-estimate of both (set both to `None`)
- `baseline`/`baseline_nonneg` (defaults `baseline_nonneg=False`, since dF/F fluctuates below zero unlike raw fluorescence)
- `penalty`: 1 for L1 (default), 0 for L0, or `None` to skip the noise-constrained optimization and deconvolve without imposing sparsity (delegates to `oasis-deconv`'s plain `oasisAR1`/`oasisAR2`)
- Always uses `roi_analyzer.sampling_frequency` (not `roi_analyzer.imaging.sampling_frequency`), consistent with `DfOverFExtension` — both work without the raw imaging loaded, matching their `need_imaging=False` declaration
- `noise_std`/`refine_kinetics`/`lam`/`s_min` available as advanced, undocumented kwargs for power users
- Parallelized per-ROI via `ProcessPoolExecutor`; `get_data(outputs='numpy'|'recording')` consistent with other extensions
- Added `oasis-deconv>=0.3.2` as a project dependency (`deconvolution` extra + `dev`)

### More realistic synthetic movies: `generate_imaging_with_rois` / `generate_fluorescence`
- Each ROI now gets its own baseline fluorescence (F0), drawn from `baseline_range`, modeling cell-to-cell brightness variability — the injected signal is scaled by each ROI's own F0, so recovered dF/F matches `clean_traces` regardless of `noise_std` or the drawn F0 values
- New `background` parameter: a dim level present everywhere in the frame, including under the ROIs (e.g. out-of-focus light), rather than only outside them — a nonzero `background` attenuates recovered dF/F the same way real neuropil/out-of-focus contamination does
- `noise_std` can be a float (fixed-variance additive Gaussian noise) or `"poisson"`, for physically realistic shot noise where variance equals the local mean signal. `background`/`baseline_range` are calibrated to realistic photon counts via a photon-transfer-curve analysis on Suite2p's own sample data (corroborated by a [GCaMP6f photon-counting blog post](https://gcamp6f.com/2016/09/20/preamplifier-bandwidth-two-ways-of-counting-photons/)); `noise_std` matches `noise_std="poisson"`'s recovered dF/F noise level via `aind_ophys_utils.signal_utils.noise_std(method="welch")`, not raw injected variance
- `event_rate` (Hz) replaces a fixed 5–15 spike count: the number of injected events per ROI now scales with the recording's duration instead of staying fixed regardless of length
- New optional `bleaching_time` passthrough to `generate_fluorescence` for photobleaching decay (renamed from `bleaching_tau`, and `decay_seconds` renamed to `decay_time`, for consistency with `DeconvolutionExtension`'s naming); `bleaching_time` defaults to `inf` (no bleaching) instead of `None`
- `generate_fluorescence`'s `traces` now consistently equals `(1 + clean_traces) * bleach(t)`, instead of special-casing no bleaching to drop the baseline offset entirely (`bleach(t) ≡ 1` is just the no-bleaching limit)
- Single-spike amplitude (dF/F peak of 1, i.e. 100%) documented as consistent with single-AP responses reported for newer, high-sensitivity indicators like jGCaMP8m/8s (Zhang et al. 2023, Nature)

### Tests
- 10 new tests for `DeconvolutionExtension`: shape/dtype/finiteness, parallel vs. serial equivalence, output types (including that the `recording` output's `sampling_frequency` matches what was actually used), ROI selection, working without loaded imaging, all decay/rise kinetics combinations (+ the invalid-input case), and `penalty=None` (plain, non-sparse deconvolution)
- 5 new tests for `generate_imaging_with_rois` (`noise_std` default/scaling, dF/F scale-independence from `noise_std`, and 2 for Poisson noise: shot-noise variance≈mean, and dF/F recovery quality at the defaults)
- Updated 1 `generate_fluorescence` test for the `traces == (1 + clean_traces)` fix

### Example
- `examples/play_with_analyzer.ipynb`: markdown section headers for each pipeline stage, a fixed seed for reproducible ground truth, noisier/more realistic synthetic data (background, per-ROI baseline, bleaching), ROI color legends, a check for ROI-to-ROI pixel overlap (unmixed crosstalk can cause small false transients), a note on background-induced dF/F attenuation, and ground-truth overlay plots (raw and rescaled) comparing inferred vs. true dF/F, denoised, and deconvolved traces

## Related upstream issues
Filed against `j-friedrich/OASIS` while building this — all closed, resolved in `oasis-deconv` 0.3.2, which this PR now depends on:
- [#32](https://github.com/j-friedrich/OASIS/issues/32) — no documented way to request AR(2) auto-estimation of both decay and rise
- [#33](https://github.com/j-friedrich/OASIS/issues/33) — support `penalty=None` for plain (non-sparse, non-noise-constrained) deconvolution, used by this PR
- [#35](https://github.com/j-friedrich/OASIS/issues/35) — `deconvolve()`: invalid `g` length silently returns `None` instead of raising
- [#36](https://github.com/j-friedrich/OASIS/issues/36) — `deconvolve()`: consider returning a NamedTuple instead of a bare 5-tuple


src/photon_mosaic/core/generators.py
```diff
@@ -228,27 +265,40 @@ def generate_imaging_with_rois(
...
+        decay_time=decay_time,
+        event_rate=event_rate,
+        bleaching_time=bleaching_time,
         seed=fluorescence_seed,
     )
-    # Inject traces into video: (T, N) @ (N, H*W*P) -> (T, H*W*P) -> (T, H, W, P)
+    # Compute the mean signal (background everywhere, plus each ROI's fluorescence trace --
+    # already F0-normalised as (1 + clean_traces) * bleach(t) -- scaled by its own F0 (baseline)):
+    # (T, N) @ (N, H*W*P) -> (T, H*W*P) -> (T, H, W, P). Scaling by F0 keeps recovered dF/F ==
+    # clean_traces regardless of noise_std, background, or the drawn per-ROI F0 values.
```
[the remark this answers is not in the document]
j-friedrich: Fixed in ebdbe01.

src/photon_mosaic/core/generators.py
```diff
@@ -261,43 +311,51 @@ def generate_fluorescence(
...
-    # Generate binary spike trains
+    # Generate binary spike trains. Each event has unit amplitude, i.e. a single, isolated
+    # spike produces a dF/F peak of 1 (100%); this is close to single-AP responses reported
+    # for newer, high-sensitivity indicators like jGCaMP8m/8s (Zhang et al. 2023, Nature).
     spikes = np.zeros((num_frames, num_rois), dtype=np.float32)
     for roi_idx in range(num_rois):
-        num_events = rng.integers(5, 15)
+        num_events = round(rng.uniform(0.5, 1.5) * event_rate * num_frames / sampling_frequency)
         event_times = rng.choice(num_frames, size=num_events, replace=False)
         spikes[event_times, roi_idx] = 1.0
```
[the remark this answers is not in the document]
j-friedrich: Fixed in ebdbe01.

src/photon_mosaic/core/tests/test_roianalyzer_core_extensions.py
```diff
@@ -372,6 +372,150 @@ def test_df_over_f_select_extension_data(analyzer_with_fluorescence, rois):
...
+    ground_truth_frames = 1000
+    ground_truth = generate_fluorescence(
+        num_frames=ground_truth_frames,
+        num_rois=NUM_ROIS,
+        sampling_frequency=SF,
+        decay_time=2.0,
+        noise_std=0.2,
+        seed=SEED,
+    )
+    analyzer_with_df_over_f.get_extension("df_over_f").data["df_over_f"] = ground_truth.traces
```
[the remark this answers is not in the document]
j-friedrich: Fixed in ebdbe01.

src/photon_mosaic/core/roianalyzer_core_extensions.py
```diff
@@ -263,6 +263,187 @@ def _select_extension_data(self, roi_ids):
...
+        if outputs == "numpy":
+            return deconvolved
+        elif outputs == "recording":
+            from spikeinterface.core import NumpyRecording
+
+            return NumpyRecording(
+                deconvolved,
+                sampling_frequency=self.roi_analyzer.sampling_frequency,
+                channel_ids=self.roi_analyzer.rois.roi_ids,
+            )
```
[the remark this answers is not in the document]
j-friedrich: Resolved in 080e06e — the sampling_frequency override was dropped entirely, so DeconvolutionExtension now always uses roi_analyzer.sampling_frequency everywhere (get_data, deconvolution computation, and the recording metadata), eliminating the possibility of disagreement.

src/photon_mosaic/core/roianalyzer_core_extensions.py
```diff
@@ -263,6 +263,181 @@ def _select_extension_data(self, roi_ids):
...
+        )
+
+    def _run(self, verbose: bool = False, **job_kwargs) -> None:
+        from concurrent.futures import ProcessPoolExecutor
+
+        dff = self.roi_analyzer.get_extension("df_over_f").get_data()
+        fs = self.roi_analyzer.sampling_frequency
+        n_jobs = fix_job_kwargs(job_kwargs).get("n_jobs", 1)
+
+        args = [(dff[:, i].copy(), fs, self.params) for i in range(dff.shape[1])]
```
alejoe91: why a `copy` here? would it be better to pass the entire `dff` buffer? then each `_deconvolve_roi` could get the index as first arg and do the slicing itself

j-friedrich: Good question — dug into this. The `.copy()` itself isn't needed: `deconvolve()` never mutates its input, and pickling a non-contiguous column slice (for the `n_jobs>1` path) already serializes just that slice's own data, not the whole `dff` matrix — verified this empirically (a column slice pickles to the same size as an explicit `.copy()`, both far smaller than pickling the full matrix). It was also being applied unconditionally, including in the `n_jobs=1` serial path where no pickling happens at all — pure wasted overhead there.

Passing the entire `dff` buffer instead (with each task doing its own slicing) would actually make the parallel case worse, though: `ProcessPoolExecutor.map()` would then pickle and send the *whole* matrix once per ROI task, rather than once per task's own slice. For a 100k-frame × 50-ROI example I tested, that's ~40MB sent 50 times instead of ~800KB × 50 (≈ the matrix's size, once total). The only way to get the benefit you're describing (share the buffer once instead of duplicating per task) would be `ProcessPoolExecutor`'s `initializer=`/`initargs=` mechanism to hand each *worker process* the array once — which is a legitimate idea, but adds real complexity (module-level global, fork/spawn-context handling) for a scenario (many ROIs, few workers) that isn't obviously the common case here. Went with the simpler fix for now: dropped the `.copy()` since it's provably unnecessary, kept the current per-task column-slicing since it's already efficient. Happy to revisit the shared-buffer approach if this becomes a real bottleneck.

Fixed in 4b5dfc8.

src/photon_mosaic/core/roianalyzer_core_extensions.py
```diff
@@ -263,6 +263,190 @@ def _select_extension_data(self, roi_ids):
...
+            False, since the input is dF/F (already baseline-subtracted and
+            free to fluctuate below zero), unlike raw fluorescence. Unused
+            when `penalty` is None.
+        penalty : int or None, optional
+            Sparsity penalty: 1 for L1 (convex), 0 for L0. If None, skips
+            the noise-constrained optimization entirely and deconvolves
+            without imposing sparsity by default (`noise_std` and
+            `baseline_nonneg` are unused in this mode). Default is 1.
+
+        Advanced, undocumented keyword arguments are also accepted and
```
j-friedrich, line 317: > **Should `DeconvolutionExtension` expose OASIS's power-user knobs at all?**
>
> It has 5 normal params (`decay_time`, `rise_time`, `baseline`, `baseline_nonneg`, `penalty`) plus 4 hidden ones passed through `**params`, undocumented in the signature: `noise_std`, `refine_kinetics`, and (only with `penalty=None`) `lam`/`s_min`. I just added validation so a typo'd kwarg raises instead of silently doing nothing, which made me wonder if we've got the right shape for these at all.
>
> **Keep hidden (status quo):** zero clutter for the common case, but not discoverable (no autocomplete, doesn't show up in `get_default_analyzer_extension_params()`).
> **Promote to the signature:** same functionality, now documented/discoverable, at the cost of a busier signature.
> **Drop entirely:** smallest surface, but power users lose the framework benefits (parallelization, caching, provenance) and have to call `oasis.functions.deconvolve()` themselves.
>
> I lean toward keeping it as-is — but curious which way you'd go.

src/photon_mosaic/core/roianalyzer_core_extensions.py
alejoe91: sounds good! Thanks for looking into this!

src/photon_mosaic/core/generators.py
```diff
@@ -228,27 +267,41 @@ def generate_imaging_with_rois(
...
-    video += (fluorescence.traces @ masks_flat).reshape(video.shape)
+    signal = roi_baseline[np.newaxis, :] * fluorescence.traces  # (T, N)
+    bleach = np.exp(-np.arange(num_frames) / (bleaching_time * sampling_frequency), dtype=np.float32)
+    mean_video = background * bleach[:, np.newaxis, np.newaxis, np.newaxis] + (signal @ masks_flat).reshape(video.shape)
+
+    noise_rng = np.random.default_rng(noise_seed)
+    if noise_std == "poisson":
+        video[:] = noise_rng.poisson(np.clip(mean_video, 0, None))
+    else:
+        video[:] = mean_video + noise_rng.normal(0, noise_std, video.shape)
```
lauraporta, line 290: These lines hold three full-size copies of the movie at once — `video`, `mean_video` (float64, promoted by `roi_baseline`), and the noise array. Peak memory is **3.06x the movie, against 2.09x on `dev`** (2000 frames, 256², float64). The notebook's `num_frames=10000` is a 4.9 GB movie; at half that it already peaks at 7.5 GB and starts swapping on a 16 GB laptop, so the full call extrapolates to ~15 GB.

Assembling it in place instead gives a **byte-identical** movie at 1.15x, and 1.5 s instead of 6.0 s:

```python
flat = video.reshape(num_frames, -1)
np.matmul(signal, masks_flat, out=flat)
flat += (background * bleach)[:, np.newaxis]
for t0 in range(0, num_frames, 256):  # noise one slab at a time
    sl = flat[t0:t0 + 256]
    sl += noise_rng.normal(0, noise_std, sl.shape)
```

I only checked the Gaussian branch, but the Poisson one can fill slab by slab the same way. `generate_random_imaging` doesn't need to change. Would that work here, or does the intermediate need to exist?

lauraporta (commented):
Very nice! Leaving one comment for now. We can chat about it, then I'm ready to approve

src/photon_mosaic/core/generators.py
j-friedrich: Confirmed the numbers and applied this — thanks for digging in.

Peak RSS for num_frames=2000, num_rois=20, 256x256 (1048.6MB movie): 3287.8MB before (3.13x, matches your 3.06x) -> 1293.7MB after (1.23x, close to your 1.15x estimate).

Assembled in-place as you suggested (matmul `out=`, in-place background add, slab-wise noise), and generalized the slab approach to the Poisson branch too, which you'd flagged as unchecked. Verified byte-identical output vs. the old implementation for the same seed, both noise_std=2.5 and noise_std="poisson" — numpy's Generator produces the same stream whether normal()/poisson() is drawn in one call or sequential same-order slabs.

Fixed in bd34838.

j-friedrich: Good catch, @lauraporta! Following up: `video` is float64 by default, so the fix above is still 1.2x of a movie that's needlessly 2x its useful size — nothing downstream reads it above float32 precision anyway.

Real 2P data is usually `uint16` anyway, not float — that'd be the bigger win (4x), but a bigger change (rounding/clipping, breaks the in-place matmul trick).

Float32 alone is straightforward and halves memory again (it did shift a seeded test's data enough to break an assumption that OASIS's serial/parallel paths are bit-exact, but that was never a real guarantee anyway, so not a big deal). Curious what you think either way.

alejoe91:
Can you clear the notebooks before merging?

j-friedrich:
Done in `f04e620` — cleared outputs/execution counts.

j-friedrich:
Recalibrated `generate_imaging_with_rois`'s defaults to realistic photon counts (`af42277`):

`background=3.0` was unrealistically bright for real 2P GCaMP imaging — corroborated by two independent sources: a [GCaMP6f photon-counting blog post](https://gcamp6f.com/2016/09/20/preamplifier-bandwidth-two-ways-of-counting-photons/) reporting 0.1-0.9 photons/pixel-bin, and a photon-transfer-curve (variance-vs-mean) analysis on Suite2p's own official sample data giving photon-count percentiles from ~0.15 (background/neuropil) to ~2 (bright cell bodies).

New defaults: `background=0.2`, `baseline_range=(0.5, 1.0)`. `noise_std=1.3` matches `noise_std="poisson"`'s recovered dF/F noise level (not raw injected variance) via `aind_ophys_utils.signal_utils.noise_std(method="welch")`.

Related to #133 (NAOMi port) — this is a lighter-weight interim improvement to the existing generator, not a replacement for that bigger effort.

j-friedrich:
This PR is now ready, @lauraporta. Generated movies now default to float32 (halves memory, no precision cost — nothing downstream reads above float32 anyway).

Also: since you commented that the data looked too bright/good/easy, I checked for realistic values for `background` and `baseline_range` (photon counts) — recalibrated both against a photon-transfer-curve analysis on Suite2p's own sample data, corroborated by a [GCaMP6f photon-counting blog post](https://gcamp6f.com/2016/09/20/preamplifier-bandwidth-two-ways-of-counting-photons/).

lauraporta (approved):
LGTM!

Sharing a plot made using this functionality. Poisson noise seems the most convincing (in line with PMT properties I would say). 

Image

Synthetic neurons overlap a bit. Maybe cool to control overlapping to test demixing in the future. (#111 CaImAn segmentation implementation)

lauraporta:
Just reduce the notebook size before merging 🙏🏻

j-friedrich:
> LGTM!
> 
> Sharing a plot made using this functionality. Poisson noise seems the most convincing (in line with PMT properties I would say).
> 
> Image
> Synthetic neurons overlap a bit. Maybe cool to control overlapping to test demixing in the future. (#111 CaImAn segmentation implementation)

Changed it in the notebook to Poisson noise, though maybe we should change the default in `generators.py` too.

Nice plots, we should put those at the end of the notebook instead of my hard-to-read ones!

PR #104, opened 2026-07-28
feat(core): support sparse ROI image masks (fixes #103)

### LINKED ISSUE #103: Use a sparse ROI representation instead of dense roi_image_masks that lead to out-of-memory errors
Currently ‎`BaseRois.get_roi_image_masks` returns an inefficient dense `np.ndarray`.
https://github.com/photon-mosaic/photon-mosaic/blob/3064b20d0e2e315c9ffc377c481d31aa8df6a42c/src/photon_mosaic/core/baserois.py#L133

Each ROI mask is also extremely sparse: a single soma or process occupies only a tiny fraction of the full imaging volume, so a dense array wastes the overwhelming majority of its allocated memory on zeros.

This will quickly lead to out-of-memory errors especially for volumetric data where there can be more ROIs than frames, i.e. representing the ROIs as a dense array would be larger than the raw data. E.g. the Suite3D [demo](https://github.com/alihaydaroglu/suite3d/blob/main/demos/02-lbm-ss004/walkthrough.ipynb) uses a video of size (1300 frames) * (22 * 929 * 736 voxels/frame) * (2 bytes/voxel) = 36.42GB (uint16), and detects 40606 ROIs which stored as dense ndarray (float32) would require (40606 ROIs) * (22 * 929 * 736 voxels/ROI) * (4 bytes/voxel) = 2.22TB.

The BaseRois representation should use a common [pydata/sparse](https://sparse.pydata.org/en/stable/) format that Suite2p and CaImAn output can be converted to, as AIND already does [here](https://github.com/AllenNeuralDynamics/aind-ophys-extraction/blob/03eafc5f84df6de301a3b409d25068dd5ff452fd/code/extraction.py#L1099) and IBL does [here](https://github.com/int-brain-lab/ibllib/blob/966b659b89d5c49119a0c1462f66355807261497/ibllib/pipes/mesoscope_tasks.py#L228).

`sparse.GCXS` is likely the better target format over `sparse.COO`: COO stores a full coordinate tuple per nonzero, while GCXS uses compressed pointer arrays (closer to scipy's CSR/CSC), which is much better suited to the matrix-multiply-heavy access pattern
this codebase actually uses (`FluorescenceNode.compute` does `chunk_flat @ masks_flat.T` on every video chunk). This also matches IBL's own precedent linked above -- it builds masks from coordinate lists as `sparse.COO` first, then explicitly converts via
`sparse.GCXS.from_coo(...)` before storing/returning them, using COO only as a construction intermediate rather than the final format.

j-friedrich:
## Summary

`BaseRois.get_roi_image_masks()` may now return a sparse [pydata/sparse](https://sparse.pydata.org/) array (e.g. `sparse.GCXS`) instead of only a dense `np.ndarray`. Each ROI typically occupies only a tiny fraction of the full imaging volume, so a dense array becomes far larger than the raw movie itself once there are many ROIs over a large (e.g. volumetric) field of view — confirmed against the numbers in #103 (2.22TB of dense masks for a 36.42GB movie).

Closes #103.

## Changes

- **`Suite2pRois`** builds masks directly as `sparse.COO` from each ROI's existing `ypix`/`xpix` coordinate lists (no dense grid ever allocated), then converts to `GCXS` for efficient downstream matrix multiplication. This is the actual real-world trigger for the OOM issue described in #103.
- **`FluorescenceNode`** and **`generate_imaging_with_rois`** use tuple-form `.reshape((n, -1))` so mask reshaping works for both dense `ndarray`s (which also accept unpacked ints) and sparse arrays (which only accept a single shape tuple). The matmul against a dense video chunk naturally returns a dense result either way — verified empirically, no further changes needed there.
- **`roianalyzer.py`**'s `create_memory` no longer wraps masks in `np.array(...)`, which raises for a sparse array instead of copying it.
- **`get_roi_pixel_masks`** uses a sparse array's own `.tocoo().coords`/`.data` directly instead of `np.nonzero`, which isn't supported on sparse arrays and would otherwise require a full densify just to find nonzero entries.
- **Binary and zarr on-disk formats** round-trip sparse masks: binary saves as `.npz` via `sparse.save_npz` (`np.save` doesn't support sparse arrays at all), zarr stores the COO `coords`/`data`/`shape` as separate datasets since zarr can't hold a sparse array as a dataset value directly.
- **`widgets/rois.py`** densifies once right after fetching masks, since rendering (matplotlib overlays, `np.where` for centroids) is inherently dense-pixel-oriented. The two backends (`plot_matplotlib`, `plot_ipywidgets`) shared this exact check, so it's factored into a small `_densify` helper.
- **`generate_rois`/`generators.py`** gained a `sparse: bool = False` option: it converts the already-built dense mask array to `sparse.GCXS` before wrapping it in `NumpyRois` (which is storage-agnostic — it just returns whatever it's given). Used by the new save/load round-trip tests' `sparse_rois` fixture.
- **`BaseRois.save(format="zarr", ...)`** works correctly through spikeinterface's unmodified `save()`/`save_to_zarr()`/`read_zarr()` chain. `_save_zarr()` writes a `zarr_class_info` attribute at the zarr root naming `ZarrRois`; `read_zarr()` checks for this attribute and reconstructs whatever class it names directly, before ever falling back to a `channel_ids`/`unit_ids` recording/sorting check that ROI data wouldn't match. `zarrimaging.py`'s `add_imaging_to_zarr_group()` uses the same mechanism for `ZarrImaging`. Because of this, `BaseRois` needs no `save()` override: `folder=` resolution (`.zarr` suffix via `clean_zarr_folder_name()`), existence checking, and `overwrite=True` support all come directly from spikeinterface's own `save_to_zarr()`.
- **`NumpyRois`** now marks itself non-JSON-serializable when backed by a sparse array: it embeds its raw `roi_image_masks` in `_kwargs` for provenance dumping, and spikeinterface's JSON encoder can't serialize `sparse.GCXS` (only dense `np.ndarray`). It now falls back to pickle provenance instead of crashing on `save(format="binary")`. Its `get_roi_image_masks()` return type annotation/docstring is also corrected from `list[np.ndarray]` to `np.ndarray | sparse.SparseArray`, matching what it actually returns.
- **All three places that build a `sparse.GCXS` mask array** (`generate_rois`, `Suite2pRois.from_stat`, `ZarrRois.get_roi_image_masks`) now pass `compressed_axes=(0,)` explicitly. `GCXS`'s default axis-compression heuristic doesn't know these arrays are indexed per-ROI (`select_rois`, `get_roi_pixel_masks(roi_ids=subset)`, `*Rois.get_roi_image_masks(roi_ids)`); benchmarked on a realistic shape (2000 ROIs, 256×256, ~5px/ROI), the default heuristic made per-ROI indexing ~40x slower than `compressed_axes=(0,)` (17ms vs. 0.45ms for a 3-of-2000 selection). This costs nothing on the bulk-matmul path in `FluorescenceNode` — after `.reshape((N, -1))` both end up compressed on axis 0 regardless, and building it that way from the start was ~2x faster there too.

## Tests

- `Suite2pRois` returns `GCXS`, and its pixel/subset/select-rois accessors handle it correctly (4 existing tests updated for sparse-aware comparisons), including an empty `roi_ids` selection.
- `test_baserois_save_load.py`: binary and zarr save/load round-trip both dense and sparse masks, via the public `.save()` API (parametrized over both).
- `FluorescenceNode` and `create_roi_analyzer` produce identical results for sparse vs. dense-backed ROIs.
- `RoisWidget`'s matplotlib backend renders correctly with sparse-backed ROIs.
- Full suite: 270 passed. Pre-commit (ruff, ruff-format, mypy, codespell): clean.

## Update: fix eager full-array loading found in review

`ZarrRois.get_roi_image_masks()` loaded the entire mask array before indexing, for both dense and sparse (indexing happened too late, and the sparse path also converted to `COO` before saving, discarding GCXS's own per-ROI `indptr` structure). Fixed: sparse now persists GCXS's own `indptr`/`indices`/`data` directly and does per-ROI lazy reads; dense restored to index the lazy zarr array before materializing (matching `dev`). Also added `Suite2pRois(sparse=False)` to opt into dense masks for small recordings.

Verified: single ROI out of 400, isolated call cost — dense 1.9MB, sparse 0.28MB (three orders of magnitude below the original 419.5MB). Sparse is also faster, not just smaller: loading all 400 ROIs is ~47x faster than dense (5.3ms vs. 253.5ms), and a single ROI ~3x faster (0.27ms vs. 0.79ms) — dense has to decompress the full array across many small per-ROI chunks regardless of what's requested.

## Known follow-up (not fixed here, tracked separately)

`RoisWidget.plot_matplotlib` densifies *all* requested ROI masks in one bulk call rather than per-ROI, which can reproduce the same class of OOM this PR fixes elsewhere if someone visualizes a very large ROI count. See #110.


src/photon_mosaic/core/numpyimaging.py
```diff
@@ -182,6 +183,9 @@ def __init__(
             "sampling_frequency": sampling_frequency,
             "roi_ids": roi_ids,
         }
+        if isinstance(roi_image_masks, sparse.SparseArray):
+            # sparse arrays aren't handled by spikeinterface's JSON encoder; fall back to pickle.
+            self._serializability["json"] = False
 
     def get_roi_image_masks(self, roi_ids: list[int | str] | None = None) -> list[np.ndarray]:
```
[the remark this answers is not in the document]
j-friedrich: Fixed in d098e85 -- updated the return type annotation and docstring to `np.ndarray | sparse.SparseArray`, matching `BaseRois.get_roi_image_masks`'s existing annotation.

j-friedrich:
Addressed the low-confidence suggestion from the [latest Copilot review](https://github.com/photon-mosaic/photon-mosaic/pull/104#pullrequestreview-4802433823): `Suite2pRois.get_roi_image_masks([])` crashed with `ValueError: At least one array required.` (from `sparse.stack([])`) instead of returning an empty array like the dense case did. Fixed in 9774cd9, with a regression test covering both `get_roi_image_masks([])` directly and via `select_rois([])`.

j-friedrich:
**Note on CI status:** the failing `tests` jobs (all platforms) are all the same single test, `test_suite2p_sample_data.py::test_suite2p_google_drive_one_file`, hitting Google Drive's "Quota exceeded" page instead of the sample TIFF it downloads over the network. This is an external rate-limit unrelated to this PR's changes — every job reports the rest of the suite passing (270/270 excluding that one test), and `linting` is green. Will re-run the failed jobs once the quota clears.

src/photon_mosaic/core/generators.py
```diff
@@ -74,6 +74,7 @@ def generate_rois(
     weighted: bool = False,
     num_planes: int = 1,
     seed: int | None = None,
+    sparse: bool = False,
```
alejoe91, line 77: should this be True by default?

j-friedrich: Kept it `False` to avoid a breaking change -- flipping it would break ~10 existing tests that call bare numpy ops (`np.unique`, `np.array_equal`, `np.argwhere`) or unpacked `.reshape(n, -1)` on the masks, none of which work on sparse. Sparse solves an OOM problem that only shows up at real segmentation scale (thousands of ROIs, large FOV) -- these synthetic fixtures stay small enough that dense is simpler and fine, real data isn't. Wanted this PR to add an opt-in, not force a rework.

src/photon_mosaic/core/zarrrois.py
```diff
@@ -65,11 +66,20 @@ def __init__(self, zarr_path, zarr_group_name: str = "rois", storage_options: di
...
+        if self._rois_group.attrs.get("roi_image_masks_sparse", False):
+            coords = np.array(self._rois_group["roi_image_masks_coords"])
+            data = np.array(self._rois_group["roi_image_masks_data"])
+            shape = tuple(self._rois_group.attrs["roi_image_masks_shape"])
+            # Compress along the ROI axis so per-ROI indexing (below, and e.g. select_rois)
+            # stays fast -- the default heuristic often picks a different axis, making it
+            # ~40x slower.
+            masks = sparse.GCXS.from_coo(sparse.COO(coords, data, shape=shape), compressed_axes=(0,))
+        else:
+            masks = np.array(self._rois_group["roi_image_masks"])
```
lauraporta, line 78: I think this is a bug. On `dev` the indexing happened first, `np.array(roi_image_masks[roi_indices])`, so zarr only read the chunks for the ROIs you actually asked for. Now the whole array gets loaded before we pick out the ones we want.

I tried it with 400 ROIs at 256×256 and asked for a single ROI: on `dev` that peaked at 1.7 MB, on this branch at 419.5 MB.

It also means the per-ROI chunking just below (L110–113) doesn't buy anything any more, since reads never go chunk by chunk. Would keeping the `np.array()` inside the indexing work here?

src/photon_mosaic/extractors/suite2prois.py
```diff
@@ -147,25 +148,43 @@ def _set_stat_properties(self, stats: Sequence[dict[str, Any]]) -> None:
...
-    def get_roi_image_masks(self, roi_ids: list[int | str] | None = None) -> NDArray:
-        """Return binary image masks shaped ``(n_rois, H, W)`` or ``(n_rois, H, W, n_planes)``."""
+    def get_roi_image_masks(self, roi_ids: list[int | str] | None = None) -> sparse.GCXS:
+        """Return binary image masks shaped ``(n_rois, H, W)`` or ``(n_rois, H, W, n_planes)``.
+
+        Returned as a sparse :class:`sparse.GCXS` array, built directly from each ROI's
+        ``ypix``/``xpix`` pixel coordinates -- suite2p detections can number in the tens of
+        thousands over large (e.g. volumetric) fields of view, where a dense array would be
+        far larger than the raw movie itself (see photon-mosaic#103).
+        """
```
lauraporta, line 158: Is sparse the only option here? As far as I can tell `Suite2pRois` always returns `GCXS`, with no way to turn it off.

It's clearly the right default, but could the choice be the user's? Something like `Suite2pRois(folder_path, sparse=False)`, for anyone whose recording is small enough that the slowdown matters more than the memory saving. (Calling `.todense()` afterwards doesn't help, since `FluorescenceNode` fetches the masks itself.)

lauraporta (changes requested):
Hey, very nice work! Good opportunity to learn about `GCXS` which I didn't know! 
I think I have found a bug, this is why I am requesting for changes. You can read it in the comments. Let me know what you think!

src/photon_mosaic/core/zarrrois.py
j-friedrich: Great catch, @lauraporta! Earlier work had the per-ROI indexing pattern in mind for sparse too — that's exactly why `compressed_axes=(0,)` was set — but didn't follow through when saving, and oddly converted to COO in the process, throwing that structure away. Confirmed, real bug, fixed.

Sparse now persists GCXS's own `indptr`/`indices`/`data` directly instead of round-tripping through COO, so a subset read only touches the requested ROIs' own slices. Dense restored to index the lazy zarr array before materializing, matching `dev`.

Verified: single ROI out of 400, isolated call cost — dense 1.9MB (matches your 1.7MB almost exactly), sparse 0.28MB — three orders of magnitude below your original 419.5MB, and still ~7x smaller than the now-fixed dense path. Also had to set an explicit chunk size for the sparse arrays along the way — zarr's default auto-chunking had no notion of the per-ROI access pattern, so a subset request was still pulling a chunk sized for the whole file.

src/photon_mosaic/extractors/suite2prois.py
j-friedrich: Agreed — added `Suite2pRois(folder_path, sparse=True)` and `Suite2pRois.from_stat(..., sparse=True)`, both defaulting to `True` (current behavior). Pass `False` for a plain dense `np.ndarray`.

src/photon_mosaic/core/zarrrois.py
j-friedrich: Follow-up: sparse isn't just smaller here, it's faster too — loading all 400 ROIs is ~47x faster than dense (5.3ms vs. 253.5ms), and a single ROI ~3x faster (0.27ms vs. 0.79ms). Dense has to decompress the full array across many small per-ROI chunks regardless of what's actually requested.

lauraporta (approved):
LGTM!

PR #118, opened 2026-08-05
Add stack_planes and split_epoch_at_frames core primitives

lauraporta:
First of two PRs splitting #83. This half is the generic `core` primitives, with no suite2p knowledge; the extractor that uses them follows in the stacked PR.

**`stack_planes`** — joins imaging objects sharing a field of view and timestamps into one multi-plane volume, lazily. Planes are pulled only from the parents that hold the requested ones, so a single plane of an N-plane volume costs one parent read rather than N (@alejoe91's review point on #83). A parent giving up all of its planes is asked with a slice, so the whole-volume read still gets views rather than advanced-indexing copies. Results are reordered to the requested order, since gathering parent by parent would otherwise return `[3, 0]` as `[0, 3]`.

Named `stack_planes` rather than `concatenate_planes` per @alejoe91: `concatenate_*` means concatenation in time, here and in SpikeInterface.

**`split_epoch_at_frames`** — subdivides one epoch of a parent into contiguous sub-epochs at the given frame boundaries, `np.split` style. Useful when an extractor exposes several recorded files as a single concatenated epoch and you want each file back as its own epoch.

Both are lazy proxies — no pixels are copied.

Tests cover plane selection across parent boundaries, requested-plane ordering, that non-contributing parents are not read at all, empty and negative selections, boolean masks, and multi-epoch. Each new test was checked to fail against the implementation it guards.

### DIFF
--- a/src/photon_mosaic/core/split.py
+++ b/src/photon_mosaic/core/split.py
@@ -1,4 +1,8 @@
-from .baseimaging import BaseImaging
+from typing import Sequence
+
+import numpy as np
+
+from .baseimaging import BaseImaging, BaseImagingEpoch
 
 
 def _normalize_epoch_indices(epoch_indices: int | list[int], num_epochs: int) -> list[int]:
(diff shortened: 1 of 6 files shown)


src/photon_mosaic/core/split.py
alejoe91, line 1: @lauraporta would it make sense to have a more general `epoch_utils/tools.py`? Then we could add here also other functions (e.g. append_epochs, concatenate, etc)

src/photon_mosaic/core/split.py
```diff
@@ -46,3 +50,84 @@ def __init__(self, imaging: BaseImaging, epoch_indices: int | list[int]):
 def split_epochs(imaging: BaseImaging, epoch_indices: int | list[int]) -> SelectEpochImaging:
     """Return a proxy imaging object with only the requested epochs."""
     return SelectEpochImaging(imaging=imaging, epoch_indices=epoch_indices)
+
+
+class _FrameRangeEpoch(BaseImagingEpoch):
```
alejoe91, line 55: Can we also make a `FrameSliceImaging(parent_imaging, start_frame, end_frame)`?

It would be super useful to cut out a stub/test imaging object out of a longer video. Implementation-wise, it would be very similar to the SplitEpochAtFramesImaging, but only with one interval.

alejoe91 (changes requested):
@lauraporta a couple of suggestions and extensions! Should be quick to address :)

lauraporta:
Hi @alejoe91, thanks for the suggestions. Happy to add `FrameSliceImaging`. Also, I am favourable on having a `epoch_utils`-like folder... Thinking of its name, I might change it to make it intuitive for internal modules containing also `Imaging` related utilities.

alejoe91:
> Hi @alejoe91, thanks for the suggestions. Happy to add `FrameSliceImaging`. Also, I am favourable on having a `epoch_utils`-like folder... Thinking of its name, I might change it to make it intuitive for internal modules containing also `Imaging` related utilities.

I would just make an `epoch_utils.py` file for now. We don't need a separate module IMO since this is only going to cover a few specific functions

lauraporta:
Both done.

I named the module `imaging_ops.py` instead of `epoch_utils.py`, since it now holds `Imaging`-level utils too, not just `Epoch` ones. I also moved `stack_planes` in, so everything that returns a lazy proxy lives in one file and `append_epochs`/`concatenate` have a home.

`FrameSliceImaging` took about 25 lines — it reuses the epoch class `split_epoch_at_frames` already needed. Unlike `FrameSliceRecording` it accepts a multi-epoch parent, but you have to say which epoch you want; see #129.

alejoe91 (approved):
LGTM!

PR #123, opened 2026-08-11
explicit delegate get_series to splitting parent

### LINKED ISSUE #120: SelectPlanesImaging bug?
Working on #118 found a possible small bug in `SelectPlanesImaging`.

I can select planes from an imaging object:
```python
img = generate_random_imaging(
    num_frames=6, height=3, width=4, sampling_frequency=10.0, num_planes=4, seed=0
)
parent_series = img.get_series()

sel = select_planes(img, [2, 3])
print(list(sel.plane_ids))
```
`[2, 3]`

If I then use `split_epochs`:
```python
epochs = split_epochs(sel, 0) 
series_from_epoch = epochs.get_series()

print(np.array_equal(series_from_epoch, parent_series[..., [2, 3]]))
```
`False`

Does not return me the series with the correct planes 2 and 3, but...
```python
print(np.array_equal(series_from_epoch, parent_series[..., [0, 1]]))
```
`True`
the planes 0 and 1.

Am I misunderstanding how to use `split_epochs`? @alessandrofelder 
Maybe the root cause could come from [`plane_indices = slice(self.get_num_planes())`](https://github.com/photon-mosaic/photon-mosaic/blob/bd6fc2eed220f6ab8f8ddfe97bea06e726052cf2/src/photon_mosaic/core/baseimaging.py#L341)

alessandrofelder:
Fixes #120 

This PR adds an explicit call to the parent imaging's `get_series` function from the `SelectEpochImaging` class (including remapping indices), because before it was ignoring any modifications to `get_series` any of its parents were doing, and going directly to the BaseImaging `get_series` implementation.

This is tested by a new test that repeatedly splits epochs.

### DIFF
--- a/src/photon_mosaic/core/split.py
+++ b/src/photon_mosaic/core/split.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from .baseimaging import BaseImaging
 
 
@@ -41,6 +43,22 @@ class SelectEpochImaging(BaseImaging):
             "imaging": imaging,
             "epoch_indices": normalized_epoch_indices,
         }
+
+    def get_series(
+        self,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        plane_ids: list | np.ndarray | None = None,
+        epoch_index: int | None = None,
+    ) -> np.ndarray:
+        # parent may want to do further operations to the series before returning
+        # so we explicitly delegate to them rather than relying on BaseImaging.
+        return self._parent.get_series(
+            start_frame=start_frame,
+            end_frame=end_frame,
+            plane_ids=plane_ids,
+            epoch_index=epoch_index,
+        )
 
 
 def split_epochs(imaging: BaseImaging, epoch_indices: int | list[int]) -> SelectEpochImaging:
--- a/src/photon_mosaic/core/split.py
+++ b/src/photon_mosaic/core/split.py
@@ -34,9 +34,11 @@ class SelectEpochImaging(BaseImaging):
         BaseImaging.__init__(self, sampling_frequency=imaging.sampling_frequency, shape=imaging.shape)
         imaging.copy_metadata(self)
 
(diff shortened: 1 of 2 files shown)


alessandrofelder:
Back into draft because tests fail (and I want to add an extra test) - sorry for the extra ping, Laura!

src/photon_mosaic/core/tests/test_split.py
```diff
@@ -28,6 +28,19 @@ def test_split_epoch_with_multiple_indices_preserves_order():
     assert split_imaging.epochs[1] is imaging.epochs[0]
 
 
+def test_split_repeatly():
```
lauraporta, line 31: Can we also include a test in which we use `select_planes` so to replicate the bug described in the issue? I can see that with this PR is actually fixed, but it is just to cover that use case.

src/photon_mosaic/core/split.py
```diff
@@ -32,16 +34,39 @@ def __init__(self, imaging: BaseImaging, epoch_indices: int | list[int]):
...
+        self,
+        start_frame: int | None = None,
+        end_frame: int | None = None,
+        plane_ids: list | np.ndarray | None = None,
+        epoch_index: int | None = None,
+    ) -> np.ndarray:
+        # parent may want to do further operations to the series before returning
+        # so we explicitly delegate to them rather than relying on BaseImaging.
+        # We also remap selected epoch indices relative to the parent
+        if epoch_index is None:
```
lauraporta, line 59: Maybe we should say in a warning say how the indexing is changing 🤔

lauraporta: I think it can be worth to expose the original indexing of the `Epochs` as we did for the planes... something like `epoch_ids`. But that's out of scope of this PR 😬

lauraporta (changes requested):

Issue #121, opened 2026-08-11
Complete `numpy`-style docstrings


alessandrofelder:
All our public functionality should have docstrings, and this should be linted using `numpydoc`.

PR #134, opened 2026-09-03
Questions for baseimaging

joannercsheppard:
Let me know if this is an ok way to do this? Please don't accept the pull request!!!

### DIFF
--- a/src/photon_mosaic/core/baseimaging.py
+++ b/src/photon_mosaic/core/baseimaging.py
@@ -37,6 +37,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
         dtype = self.get_dtype()
         sf_hz = self.sampling_frequency
 
+        # why is this in-line when convert_seconds_to_str and convert_bytes_to_str are used below? It seems like it could be a separate function for clarity.
         # Format sampling frequency
         if not sf_hz.is_integer():
             sampling_frequency_repr = f"{sf_hz:f} Hz"
@@ -74,9 +75,12 @@ class BaseImaging(BaseExtractor, TimeSeries):
         )
 
     def __repr__(self):
+        # missing docstring
+        # also why needed? this might just be me being dumb about good practices, but it seems like this is just calling _repr_header, which is already called in _repr_html_. Is there a reason to have both?
         return self._repr_header()
 
     def _repr_html_(self, display_name=True):
+        # missing docstring
         common_style = "margin-left: 10px;"
         border_style = "border:1px solid #ddd; padding:10px;"
 
@@ -176,9 +180,11 @@ class BaseImaging(BaseExtractor, TimeSeries):
         self.add_segment(epoch)
 
     def get_sampling_frequency(self):
+        # missing docstring
         return self._sampling_frequency
 
     def get_sample_size_in_bytes(self):
+        # missing docstring
         return self.get_num_pixels() * np.dtype(self.get_dtype()).itemsize
 
     def get_shape(self, segment_index: int | None = None) -> tuple:
@@ -253,6 +259,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
         int
             The total number of frames.
         """
+        # why is this called get_num_frames when the other function is called get_num_samples? Is this a byproduct of this coming from spikeinterface? If so, it might be worth adding a comment to clarify that.
         return self.get_num_samples(segment_index=epoch_index)
 
     def get_total_frames(self) -> int:
--- a/src/photon_mosaic/core/baserois.py
+++ b/src/photon_mosaic/core/baserois.py
@@ -42,6 +42,7 @@ class BaseRois(BaseExtractor):
         return f"{name}:\n{self.get_num_rois()} ROIs - {shape_repr}"
 
     def _repr_html_(self, display_name=True):
+        # no docstring
         common_style = "margin-left: 10px;"
         border_style = "border:1px solid #ddd; padding:10px;"
 
@@ -89,6 +90,7 @@ class BaseRois(BaseExtractor):
 
     @property
     def sampling_frequency(self):
+        # no docstring
         return self._sampling_frequency
 
     @property
@@ -102,6 +104,7 @@ class BaseRois(BaseExtractor):
         """
         return self._roi_ids
 
+    # super pedantic but in baseimaging.py all the @property methods are grouped together but here there interlaced - should this be consistent? (i.e. all @property methods together and all other methods together)
     def get_num_planes(self) -> int:
         """Get the number of planes.
 
@@ -198,7 +201,7 @@ class BaseRois(BaseExtractor):
 
     def register_imaging(self, imaging: BaseImaging):
         """
-        Register an imaging to the ROIs. If the ROIs and imaging both contain
+        Register an 'imaging object' or 'image'?? to the ROIs. If the ROIs and imaging (as before) both contain
         time information, the imaging's time information will be used.
 
         Parameters
--- a/src/photon_mosaic/core/binaryimaging.py
+++ b/src/photon_mosaic/core/binaryimaging.py
@@ -88,9 +88,11 @@ class BinaryImaging(BaseImaging):
         }
 
     def is_binary_compatible(self) -> bool:
+        # docstring missing
         return True
 
     def get_binary_description(self):
+        # docstring missing
         d = dict(
             file_paths=self._kwargs["file_paths"],
             dtype=np.dtype(self._kwargs["dtype"]),
@@ -138,6 +140,8 @@ class BinaryImagingEpoch(BaseImagingEpoch):
         end_frame: int,
         plane_indices: slice | np.ndarray | None = None,
     ) -> np.ndarray:
+        # docstring missing
+
         # Calculate byte offsets for start and end frames
         start_byte = self.file_offset + start_frame * self.bytes_per_sample
         end_byte = self.file_offset + end_frame * self.bytes_per_sample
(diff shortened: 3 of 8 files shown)


src/photon_mosaic/core/baseimaging.py
```diff
@@ -253,6 +259,7 @@ def get_num_frames(self, epoch_index: int | None = None) -> int:
         int
             The total number of frames.
         """
+        # why is this called get_num_frames when the other function is called get_num_samples? Is this a byproduct of this coming from spikeinterface? If so, it might be worth adding a comment to clarify that.
```
alessandrofelder, line 262: I think this is a byproduct of coming via SpikeInterface. @alejoe91 can maybe confirm?
And I agree about adding the docstring.

src/photon_mosaic/core/baseimaging.py
```diff
@@ -74,9 +75,12 @@ def _repr_header(self, display_name=True):
         )
 
     def __repr__(self):
+        # missing docstring
+        # also why needed? this might just be me being dumb about good practices, but it seems like this is just calling _repr_header, which is already called in _repr_html_. Is there a reason to have both?
```
alessandrofelder, line 79: I think this is needed to cover the use case where we want to print an `Imaging` object outside of `ipython` (e.g. `print(imaging)` in a Python script)

Docs I consulted to inform my thoughts:
* https://docs.python.org/3/library/functions.html#repr
* https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

And then, they both call our own (AFAICT?) `_repr_header_` to avoid code repetition?

100% docstrings would help!

src/photon_mosaic/core/baseimaging.py
```diff
@@ -37,6 +37,7 @@ def _repr_header(self, display_name=True):
         dtype = self.get_dtype()
         sf_hz = self.sampling_frequency
 
+        # why is this in-line when convert_seconds_to_str and convert_bytes_to_str are used below? It seems like it could be a separate function for clarity.
```
alessandrofelder, line 40: I guess it doesn't exist in SpikeInterface's `core_tools` (but I haven't checked)... it should probably go there if we want to make it a function?

alessandrofelder (commented):
Thanks @joannercsheppard - I generally agree with you that more docstrings would be nice, even in somewhat trivial functions, especially for such a central class and with a view that we would like a Reference API docs based on docstrings on our website.

Related to #121

src/photon_mosaic/core/baseimaging.py
lauraporta: Guess this should become `convert_hertz_to_str` 🤔
If it is useful to spikeinterface as well I agree it should be added to the upstream class

src/photon_mosaic/core/baseimaging.py
lauraporta: I would agree with Ale, it seems it is to catch both the `__repr__` built in and the signature expected by rich formatting... 🤔

src/photon_mosaic/core/baseimaging.py
```diff
@@ -74,9 +75,12 @@ def _repr_header(self, display_name=True):
         )
 
     def __repr__(self):
+        # missing docstring
+        # also why needed? this might just be me being dumb about good practices, but it seems like this is just calling _repr_header, which is already called in _repr_html_. Is there a reason to have both?
         return self._repr_header()
 
     def _repr_html_(self, display_name=True):
+        # missing docstring
```
lauraporta, line 83: Plenty of these cases 😬

src/photon_mosaic/core/baseimaging.py
lauraporta: Is there possibly a difference in which samples is across all the `Epochs`? And frames within one `Epoch`?

alessandrofelder: Both take an index in the same way, and one calls the other, so I think this is just a wrapper with a more imaging-related name allowing us to pass `epoch_index` instead of `segment_index`

lauraporta: Alright so basically duplicated

alessandrofelder: Yea just translates to `photon-mosaic` terminology.

src/photon_mosaic/core/baseimaging.py
joannercsheppard: Oh yeah I didn't spot that it was importing from spikeinterface!

joannercsheppard:
FYI just committed rest of questions from core but idk if you will both get notifications @lauraporta @alessandrofelder

src/photon_mosaic/core/baserois.py
```diff
@@ -102,6 +104,7 @@ def roi_ids(self) -> np.ndarray:
         """
         return self._roi_ids
 
+    # super pedantic but in baseimaging.py all the @property methods are grouped together but here there interlaced - should this be consistent? (i.e. all @property methods together and all other methods together)
```
alessandrofelder, line 107: I think this is a nice idea? Feel free to open an issue... not sure if there's a linting rule we could stick in a .precommit.yaml for this 🤔

src/photon_mosaic/core/baserois.py
```diff
@@ -198,7 +201,7 @@ def select_rois(self, roi_ids: ArrayLike) -> "BaseRois":
 
     def register_imaging(self, imaging: BaseImaging):
         """
-        Register an imaging to the ROIs. If the ROIs and imaging both contain
+        Register an 'imaging object' or 'image'?? to the ROIs. If the ROIs and imaging (as before) both contain
```
alessandrofelder, line 204: Probably 'Imaging object`? or just capital `Imaging`?
I agree it's a bit of a weird word, it makes sense to me as a thing inbetween and image and a recording but we are stretching the language a bit. Are you suggesting we revisit the term, or do you just think that the docstring is unclear?

src/photon_mosaic/core/binaryimaging.py
```diff
@@ -241,6 +245,7 @@ def __init__(self, folder_path):
         self._kwargs = dict(folder_path=str(Path(folder_path).absolute()))
         self._bin_kwargs = d["kwargs"]
 
+    # seems like a repeat to above is_binary_compatible and get_binary_description methods, but for folders? feels like there might be way to combine? or maybe naming should be slightly different to avoid confusion?
```
alessandrofelder, line 248: We could experiment and implement in a branch to evaluate what the trade-off in terms of added complexity (presumably some extra if/else clauses) versus avoiding code duplication?

src/photon_mosaic/core/imaging_tools.py
```diff
@@ -2,7 +2,7 @@
 
 from .baseimaging import BaseImaging
 
-
+# noticing that this file has a lot of defined types which looks different to the other files. Is this a style choice? If so, should it be consistent across all files? (i.e. either use defined types or not use defined types)
```
alessandrofelder, line 5: Can you make an example of what you mean by defined types here? (Typehints?)

We should definitely be stylistically consistent across files (and I would vote for using typehints if that's what you mean).

src/photon_mosaic/core/roianalyzer_core_extensions.py
```diff
@@ -345,6 +346,20 @@ def _kde_mode_percentile(data: np.ndarray, N: int = 2**12) -> float:
...
+        ----------
+        t : float
+            Current estimate of the bandwidth.
+
+        Returns
+        -------
+        float
+            Difference between the current estimate and the updated estimate.
+        ''' 
+        # wasn't sure what this did but this is what Claude auto-completed. does it make sense??
```
alessandrofelder, line 362: This is one for @j-friedrich 😅 (or for us to discuss when reading the corresponding paper together 😁 )

src/photon_mosaic/core/roianalyzer.py
```diff
@@ -294,6 +294,27 @@ def create_memory(cls, rois, imaging, imaging_attributes):
...
+            Attributes describing the imaging data.
+        backend_options : dict
+            Additional options for the backend.
+
+        Returns
+        -------
+        RoiAnalyzer
+            An instance of RoiAnalyzer configured to use the "binary_folder" backend.
+        '''
+        # once again just autocompleting w Claude/
```
alessandrofelder, line 317: Looks reasonable to me 🤷

src/photon_mosaic/core/roianalyzer_core_extensions.py
```diff
@@ -118,6 +118,7 @@ def __init__(
         else:
             self._neuropil_flat = None
 
+    # why not just return float32 directly for consistency with the compute method?
```
alessandrofelder, line 121: Not sure I understand this?

src/photon_mosaic/core/selectplanes.py
```diff
@@ -1,3 +1,6 @@
+'''
```
alessandrofelder, line 1: I think these are good docstrings.
We should have a linter for whether docstrings use 3x" or 3x' 😅

alessandrofelder (commented):
Possibly useful thoughts on the second commit 🤞 thanks @joannercsheppard 

Feels like generally, it's time to tidy up our experiments as we work towards a first release and implement docstrings and typehints etc.

lauraporta:
@joannercsheppard, would it be OK to close this PR?

src/photon_mosaic/core/baseimaging.py
alejoe91: yes just a translation to bring it to the imaging world ;)

src/photon_mosaic/core/baseimaging.py
alejoe91: do we need docstring here? This is a magic function for IPython to render in html in jupyter

src/photon_mosaic/core/baserois.py
alejoe91: Yes this should say `BaseImaging`

Issue #132, opened 2026-09-02
BasePreprocessor dtype in constructor unused?


alessandrofelder:
https://github.com/photon-mosaic/photon-mosaic/blob/43cfc029f16639b7a9a8258207e5798803c3486f/src/photon_mosaic/preprocessing/basepreprocessor.py#L10 takes an optional `dtype` argument but it just gets set as a local variable inside the constructor and is then forgotten? Or maybe I don't understand the intention / design here?

alejoe91:
Good catch. This is a mismatch with the SpikeInterface API. In spikeinterface, the `BaseRecording` expects a `dtype` as input, which is stored internally so that the Base object knows its dtype: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/preprocessing/basepreprocessor.py#L20

So 2 optiosn here: 

1. extend the `BaseImaging` object to do the same. This would allow us to avoid "sampling data" when retrieving the dtype (see [here](https://github.com/photon-mosaic/photon-mosaic/blob/dev/src/photon_mosaic/core/baseimaging.py#L278-L286))

2. keep it as is and remove the `dtype` form the `BasePreprocessor`

I vote for option 1. @photon-mosaic/active-devs thoughts?

PR #138, opened 2026-09-08
Extending BaseImaging to track dtype to match SpikeInterface (#132)

joannercsheppard:
BaseImaging extended to store dtype instead of inferring from samples so now BaseProcessor uses the dtype in constructor.

### DIFF
--- a/src/photon_mosaic/core/baseimaging.py
+++ b/src/photon_mosaic/core/baseimaging.py
@@ -19,7 +19,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
     The `_main_ids` attribute is used here for multi-plane imaging objects.
     """
 
-    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike):
+    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike, dtype: DTypeLike | None = None):
         # Should we allow users to provide 2D shape (H, W) for single plane imaging?
         if len(shape) == 2:
             shape = (shape[0], shape[1], 1)
@@ -29,6 +29,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
         self._sampling_frequency = float(sampling_frequency)
         self._shape = tuple(shape)  # Image is intended as a volume (H, W, planes)
         self._average_image = None
+        self._dtype = np.dtype(dtype) if dtype is not None else None
 
     def _repr_header(self, display_name=True):
         """Generate text representation of the BaseImaging object."""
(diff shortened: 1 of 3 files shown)


src/photon_mosaic/core/baseimaging.py
```diff
@@ -19,7 +19,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
     The `_main_ids` attribute is used here for multi-plane imaging objects.
     """
 
-    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike):
+    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike, dtype: DTypeLike | None = None):
```
alejoe91, line 22:
```suggestion
    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike, dtype: DTypeLike):
```
I think we should enforce `dtype` to be not `None`

alejoe91:
@joannercsheppard can you fix linting test? My suggested change means that we have to add `dtype` to a bunch of places :)

src/photon_mosaic/core/baseimaging.py
```diff
@@ -19,7 +19,7 @@ class BaseImaging(BaseExtractor, TimeSeries):
     The `_main_ids` attribute is used here for multi-plane imaging objects.
     """
 
-    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike):
+    def __init__(self, sampling_frequency: float, shape: tuple | list | ArrayLike, dtype: DTypeLike):
```
alessandrofelder: Should `dtype` be `None` by default or do we expect users to explicitly set it to `None` if they want to infer from data? Maybe I misunderstood 🤔

src/photon_mosaic/preprocessing/tests/test_basepreprocessor.py
```diff
@@ -37,6 +38,24 @@ def test_init_custom_dtype(self, imaging):
...
+        assert reg.get_dtype() == np.dtype(np.float32)
+
+    def test_get_dtype_defaults_to_parent_dtype(self, imaging):
+        # Verify that the default dtype is inherited from the parent imaging object
+        reg = BasePreprocessor(imaging)
+        assert reg.get_dtype() == imaging.get_dtype()
+
+    def test_get_dtype_with_explicit_dtype_does_not_sample_parent_data(self):
+        # An imaging object with no epochs has no data to sample. If get_dtype()
+        # tried to infer the dtype from data, this would raise; an explicit
```
alessandrofelder: Nice edge case! I think the case mentioned here currently raises with a cryptic error message (an `IndexError`) - maybe we want to catch that an return a more descriptive error message along the lines of "empty imaging objects require an explicit datatype"?

PR #144, opened 2026-09-14
Make ImagingSeriesWidget playback thread-safe

j-friedrich:
**TL;DR:** `_playback_loop` runs in a background thread sharing mutable
state with several UI callbacks (play/pause, seek, FPS) with no
synchronization. Adds one lock making every transition atomic, a wake
event so those callbacks take effect immediately instead of waiting out
a poll interval, and a generation counter so frame publishing/rendering
can't tear or go stale. ~24 distinct races/bugs fixed. Stacked on #141.

## Summary
Stacked on top of #141 (base branch: `fix-imaging-series-playback`, not
`dev` -- depends on that PR's fixes, retarget once it merges). `68d2223`
below re-fixes the same final-frame-stop-delay bug #141 fixes separately
as `e675024` -- expect a trivial same-line conflict on rebase.

`ImagingSeriesWidget` playback (`_playback_loop`, its own thread) shares
mutable state -- `current_frame`, `_playback_last_time`, `is_playing`,
`play_thread` -- with the FPS slider, frame slider, play/pause button,
and `seek_to_frame`, all of which can run concurrently with it. None of
that access was synchronized.

## Changes
Three groups of fixes, all under one new lock (`_playback_timing_lock`):

- **State transitions made atomic**: play/pause/start/stop, the
  end-of-playback decision, mid-shutdown restarts, and a
  `_playback_start_id` generation counter so a stale "reached the end"
  decision can't cancel a fresh restart.
- **A wake event replaces polling `time.sleep`**: starting, seeking,
  changing FPS, and stopping all now interrupt a sleeping worker
  immediately instead of leaving it to notice on its own next poll (up
  to 2.5s at the slowest FPS) -- plus several edge cases in exactly when
  to clear/set that event so a wake-up landing in a gap can't get
  silently dropped.
- **Frame publishing/rendering made race-safe**: a `_frame_generation`
  counter lets a publish or a seek converge on the true state instead of
  leaving the slider stale or torn; a separate `_render_lock` (guarding
  only the Matplotlib mutations, not the slow frame fetch) stops two
  concurrent renders from interleaving into a torn frame; display
  settings (vmin/vmax/colormap) are read consistently once per render
  rather than per view.

Each fix's specific race and reasoning lives in its own commit message
and in a comment at that fix's exact location in `series.py` -- this
description gives the shape of the change, not a bullet per fix.

## Tests
A regression test per fix, each confirmed to fail against the code it
fixes -- several force genuine thread interleaving (real
`threading.Thread` + blocking-lock harnesses) rather than relying on
timing. A few review rounds also caught test-quality issues in the
harnesses themselves (a non-deterministic proof rewritten to be direct,
fake locks miscalibrated after production code shifted, one stale
docstring) -- none of which had ever failed in CI. Recently consolidated
several near-duplicate tests via `pytest.mark.parametrize`, removing
~235 lines with no coverage change.

### DIFF
--- a/src/photon_mosaic/widgets/series.py
+++ b/src/photon_mosaic/widgets/series.py
@@ -376,17 +376,34 @@ class ImagingSeriesWidget(BaseWidget):
         self.play_button.button_style = "success"
 
     def _playback_loop(self):
-        """Main playback loop running in separate thread."""
+        """Main playback loop running in separate thread.
+
+        Paced by elapsed wall-clock time rather than a fixed per-iteration increment: each
+        redraw involves a full figure re-render plus a Jupyter comm/websocket round trip, which
+        can take longer than ``1 / playback_fps``. A naive ``current_frame += 1`` on a fixed
+        timer would race ahead of what's actually reached the browser -- since ipywidgets only
+        syncs the latest value, whichever intermediate frames never got flushed are silently
+        dropped, with no control over which ones. Instead, each iteration computes how many
+        frames *should* have elapsed since the last actual advance and jumps straight there --
+        deliberately skipping frames (evenly, by real elapsed time) so playback speed stays
+        correct under load, rather than an uncontrolled, backpressure-dependent frame drop. The
+        reference timestamp only moves forward on an actual advance (not every poll), so this
+        also stays correct if ``playback_fps`` changes mid-playback via the FPS slider.
+        """
         import time
 
         dp = to_attr(self.data_plot)
 
+        last_time = time.monotonic()
         while self.is_playing and self.current_frame < dp.num_frames - 1:
-            time.sleep(1.0 / self.playback_fps)
-            if self.is_playing:  # Check again in case it was stopped
-                self.current_frame += 1
+            now = time.monotonic()
+            frames_elapsed = int((now - last_time) * self.playback_fps)
+            if self.is_playing and frames_elapsed > 0:  # Check again in case it was stopped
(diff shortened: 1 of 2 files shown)


j-friedrich:
Following up on the CI-time/testing-level discussion on #141 ([here](https://github.com/photon-mosaic/photon-mosaic/pull/141#discussion_r4027973675)) — the same breakdown applies more to this PR.

The ~30 concurrency tests split into two tiers:

- **User-reachable races** — play/pause double-clicks, seeking during playback, dragging the FPS slider mid-playback. Not hypothetical: the slider-drag bug found by manual testing (#146) is a sibling of this exact category.
- **Narrow internal race-windows** — e.g. `test_playback_loop_uses_fresh_state_at_each_checkpoint`, `test_playback_loop_does_not_cancel_a_restart_that_races_its_own_end_decision`, `test_publish_does_not_finalize_stale_state_while_a_seek_is_mid_flight`, the render-tear tests. These require an action landing in a specific sub-millisecond gap between two lock acquisitions — no human clicking a mouse will realistically land there.

Worth being honest about how this PR got this large: it grew entirely through an iterative Copilot-review-and-fix loop on #141, not an upfront decision that playback needed to be thread-safe. That scale of iteration — 24 distinct races found, fixed, and regression-tested — is realistically only practical with LLM assistance; without it, this PR likely wouldn't exist in this form, or at all. Worth deciding deliberately whether the narrow-race-window tier is worth keeping at that cost.

Issue #102, opened 2026-07-15
Rename `compute_motion_suite2p` into general `compute_motion`


alejoe91:
Since we plan more methods, can we rename to something like:

```
import photon_mosaic.preprocessing as ppre

motion_suite2p = ppre.compute_motion(imaging, method="suite2p", **params)
```

Same holds for `register_motion_suite2p`, but we need to figure out how general that function is, i.e., can i run from a `core.Motion` object or does it need a `Suite2pMotion`?

E.g.
```
imaging_registered = ppre.register_motion(imaging, motion=motion_suite2p, method="something")
```

lauraporta:
Related to #89

lauraporta:
## Generalising the registration classes

Today `suite2p_registration.py` defines five backend-specific surfaces (settings, motion subclass, `compute_motion_suite2p`, `RegisterSuite2PImaging` + `RegisterSuite2PImagingEpoch`, `register_suite2p`), and #89 would require abstractions in order to avoid duplication.

Suggestions with inheritance in parenthesis:
| today (per backend) | proposed | 
|---|---|
| `Suite2pRegistrationSettings(BasePreprocessor)` | `RegistrationSettings(BasePreprocessor)` + `Suite2pRegistrationSettings(RegistrationSettings)` + `JNormcorreRegistrationSettings(RegistrationSettings)` |
| `core.Motion`, `Suite2PMotion(Motion)` | `core.Motion` + `Suite2PMotion(Motion)` + `JNormcorreMotion(Motion)` | 
| `compute_motion_suite2p()` | `compute_motion(imaging, method, settings, **params) -> Motion`; should it become a class and containing `_compute_reference_wrapper()` as a method? That's the only function using it...|
| `RegisterSuite2PImaging` + `RegisterSuite2PImagingEpoch` | `RegisterImaging(BasePreprocessor)` + `RegisterSuite2PImagingEpoch` + `RegisterJNormcorreImagingEpoch`; I would not have an abstract RegisterImagingEpoch as it's mostly `get_series()` and it is going to be totally different between the two |

Let me know what do you think. Also unsure if to make two separate modules for settings, motion and registration...

arielleleon:
Thanks for outlining this. I don't think I follow row two of your proposed table. I think we are thinking the same thing but want to make sure. In your table, I wasn't sure about where some of the classes live but that's probably a me issue. Let me know if this resonates

```
compute_motion(imaging, method=...)        <- shared
[...]

PR #148, opened 2026-09-18
refactor(preprocessing): backend-agnostic compute_motion and Register Imaging (#102)

lauraporta:
HI @arielleleon, here's a possible implementation of what we have discussed, what do you think?

---
Motion gains a compute() classmethod and a lazy method registry, so backends are selected by name and imported only when used. Shared RegistrationSettings and RegisterImaging move to preprocessing/registration.py; suite2p re-parents onto them. Old names kept as aliases.


src/photon_mosaic/core/motion.py
```diff
@@ -1,10 +1,98 @@
...
 from numpy.typing import NDArray
 
 from photon_mosaic.core import BaseImaging
 
+# Method name -> module that defines (and registers) the backend's Motion
+# subclass. Only module *paths* live here, so core imports no backend: the
+# module is imported lazily the first time the method is requested.
+_builtin_motion_modules: dict[str, str] = {
+    "suite2p": "photon_mosaic.preprocessing.suite2p_registration",
+    "jnormcorre": "photon_mosaic.preprocessing.jnormcorre_registration",
```
alejoe91: can't this be filled by the `register_motion_class` too?

src/photon_mosaic/core/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .binaryrois import BinaryFolderRois, BinaryRois
 from .generators import generate_fluorescence, generate_imaging_with_rois, generate_random_imaging, generate_rois
 from .loading import load
-from .motion import Motion
+from .motion import Motion, coerce_settings, compute_motion, register_motion_class
```
alejoe91:
```suggestion
from .motion import Motion, compute_motion, register_motion_class
```
do end users need `coerce_settings`?

alejoe91 (commented):
Couple of small comments. I'll give more feedback by next week

src/photon_mosaic/core/motion.py
arielleleon: I don't see how that would help us achieve what `_builtin_motion_modules` is trying to achieve which is to tell `core` which module to import. In this case, `register_motion_class` runs only as a side effect of importing the backend module. If we implement `register_motion_class` alone, then we would have to import all registration methods.

src/photon_mosaic/core/__init__.py
arielleleon: No, should be internal only

